PR #9, opened 2015-06-08
Add progress bar to pandas apply function

hadim:
Tell me if you are ok to include this.


hadim:
The following include a big copy-paste code from `_tqdm.py`. I really need to find a way to avoid this code duplication.

kmike:
I like this feature (wanted a tqdm for DataFrame.apply myself). 
+1 to find a way to reduce code duplication. 
It'd be also nice to mention this feature in README.

hadim:
Cool !

I will work on it next week.

tqdm/tests/tests_pandas.py
```diff
@@ -0,0 +1,54 @@
...
+from nose.plugins.skip import SkipTest
+from nose.tools import with_setup
+
+
+def setup_pandas():
+    try:
+        from tqdm import enable_progress_apply
+        enable_progress_apply()
+    except:
+        raise SkipTest
```
kmike: To run this test on Travis pandas should be added to tox.ini

lrq3000:
This should be updated with the new object oriented architecture of tqdm, and merged! This is a great addition.

hadim:
Sure but I don't have time right now to dig into this. Maybe later !

lrq3000:
Yes I understand, it's not urgent, but it sure is a very good added value :D I will myself use that in the future without a doubt

casperdcl:
Having never used `pandas` due to its various inefficiencies, I'm not sure if this is what you guys wanted. But I tried...

casperdcl:
I don't know if you'd like me to revert back to `tqdm_pandas(tqdm(**kwargs))` instead of `tqdm_pandas(**kwargs)` since the former allows for `tqdm_gui` and other derived classes to be used instead...

lrq3000:
@casperdcl : `tqdm_pandas(**kwargs)` is simpler, but since we started to modularize the code, I think it would be more logical to revert to `tqdm_pandas(tqdm(**kwargs))`, so that future derived classes will be useable without breaking compatibility. Thank's for the work :+1: 

I don't use pandas currently, can someone that use pandas extensively test this out?

casperdcl:
The rewritten tests seem to work... Travis seems to be taking its time to install `numpy` and `pandas`, though.

casperdcl:
Also todo: answer the following to gain popularity...

- [ ] http://stackoverflow.com/questions/18603270/progress-indicator-during-pandas-operations-python
- [ ] http://stackoverflow.com/questions/24737683/how-do-i-create-a-progress-bar-when-a-dataframe-is-initializing

casperdcl:
@hadim @kmike can you check if this PR makes your day ;) ?

hadim:
Will do this week.

hadim:
Working nicely to me. Very nice work @casperdcl !

Good to merge for me.

casperdcl:
required surprisingly few loc, actually... kind of like the rest of `tqdm`. Most of our repo is documentation and tests :) ... it's really a useful template for other projects, hah

lrq3000:
Totally agree! :+1:

2015-12-01 0:55 GMT+01:00 Casper da Costa-Luis <notifications@github.com>:

manugarri:
This is way cool, looking forward to see it on a release!

casperdcl:
@manugarri it's been released for ages... try `pip install -U tqdm` if you need to upgrade... Or use the version from github.

PR #16, opened 2015-06-18
Add flag to travis file to switch to container based infrastructure

hadim:
It should allow faster build.


hadim:
Easy merge so I took it :-)

Issue #50, opened 2015-11-10
jitter in IPython notebook


kmike:
Hey,

This is how tqdm 2.2.4 looks in IPython notebook: http://www.youtube.com/watch?v=GjRHAmj_xfc

This is how it used to look in 1.0: http://www.youtube.com/watch?v=t7e6IEdEaTc

Red color is fine (we're using stderr now, and it makes sense), but I think text shouldn't jump by default, and progress bar looks untidy (it is not image compression artifact - there are vertical lines, and the last symbol doesn't have the same height as other symbols).

I also find old `elapsed: ... left: ...` to be easier to understand than new `...<...`, but that's a personal preference :)

Code:

```py
import time
import random
from tqdm import tqdm

x = range(343)

for el in tqdm(x, "Loading", mininterval=0, leave=True):
    time.sleep(random.random() * 0.01)
```

lrq3000:
Hey @kmike, what version of IPython are you using? Did you try the latest Jupyter version?

I have tried on my Jupyter notebook, and there's no issue like the one we can saw on the video: the only moment the bar is momentarily jumpy is when the number go past a new magnitude (eg, 100 -> 1000), but this also happened before. On my notebook, the smooth (unicode) bar does progress correctly without causing any issue. Indeed the small blocks are also smaller in height and not only width, but they are centered vertically so that's fine (contrary to your install where they are not centered).

Here's a screenshot of what I see:
![jupyter-notebook-tqdm](https://cloud.githubusercontent.com/assets/1118942/11073179/a6b1b0a8-87e9-11e5-9e71-0bb0af5dd630.png)

Also, you can probably get the same result as tqdm v1.0 by setting ascii=False, since I guess the issue is in the management of unicode characters in your IPython install.

kmike:
Yeah, I know `ascii` argument may fix it, but this doesn't look like a good default. 
I'm using latest Jupyter notebook, Python 3.4, Firefox 42.0, OS X. The issue is more visible with small `mininterval` values.

lrq3000:
@kmike : correct, I can confirm the issue when `mininterval=0` on Chrome browser, Windows 7. And I can confirm that `ascii=True` fix the issue just like with tqdm v1.0.

So it seems that a few blocks are doing weird things because of their size variation. Unluckily, I'm not sure we can correct that in an easy way. Maybe we should just remove those buggy blocks? The progress bar will be less smooth, but also less jumpy.

kmike:
[...]

PR #61, opened 2015-11-30
Add ipython/jupyter notebook support

lrq3000:
Support for IPython/Jupyter notebook progress bar (fix for #50).

This should work on all IPython versions from v2-dev, but I only tested on the latest (Jupyter v4), but now the widgets API should be stable (this was not the case for IPython 3, there was an "experimental warning" message). Here's what it looks like:

tqdm-ipython

Also, this supports inner progress bars up to any level:

```
from tqdm import tqdm_notebook, tnrange
from time import sleep
for i in tnrange(10, desc='1st loop', leave=True):
    for j in tnrange(100, desc='2nd loop'):
        sleep(0.01)
```

Result:
tqdm-inner-pbar

To do it as cleanly as possible, I had to decouple format_meter() from tqdm() and moved it to StatusPrinter(). This will probably benefit to future derived classes (because they can define their own StatusPrinter() or format_meter() without having to redefine all tqdm functions). I also added a `nobar=False` argument to format_meter() to return only status indicators without the bar.

I didn't test the performance hit, I hope it's not significant. This should be assessed before merging.

Todo list (please help with your ideas! :D):
- [x] perf test if base tqdm() still has the same speed (after format_meter() decoupling)
- [x] perf quantify how much tqdm_notebook() is slower compares to base tqdm().
- [x] select one of the two bar designs: using ipywidgets or using custom HTML+JS+CSS ?
- [ ] coverage?
- [ ] add to readme
- [ ] bump version

### DIFF
--- a/tqdm/__init__.py
+++ b/tqdm/__init__.py
@@ -5,7 +5,9 @@ from ._tqdm import format_meter
 from ._tqdm_gui import tqdm_gui
 from ._tqdm_gui import tgrange
 from ._tqdm_pandas import tqdm_pandas
+from ._tqdm_notebook import tqdm_notebook
+from ._tqdm_notebook import tnrange
 from ._version import __version__  # NOQA
 
-__all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange', 'format_interval',
-           'format_meter', 'tqdm_pandas', '__version__']
+__all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange', 'tqdm_notebook', 'tnrange',
+           'format_interval', 'format_meter', 'tqdm_pandas', '__version__']
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -70,7 +70,7 @@ def format_interval(t):
 
 
 def format_meter(n, total, elapsed, ncols=None, prefix='', ascii=False,
-                 unit='it', unit_scale=False, rate=None):
+                 unit='it', unit_scale=False, rate=None, notebook=False, nobar=False):
     """
     Return a string-based progress bar given some parameters
 
@@ -174,7 +174,10 @@ def format_meter(n, total, elapsed, ncols=None, prefix='', ascii=False,
             full_bar = bar + \
                 ' ' * max(N_BARS - bar_length, 0)
 
-        return l_bar + full_bar + r_bar
+        if nobar:
+            return l_bar[:-1] + ' ' + r_bar
+        else:
+            return l_bar + full_bar + r_bar
 
     # no total: no progressbar, ETA, just progress stats
     else:
@@ -194,7 +197,8 @@ def StatusPrinter(file):
 
     last_printed_len = [0]  # closure over mutable variable (fast)
 
-    def print_status(s):
+    def print_status(*args, **kwargs):
+        s = '' if args[0] is None else format_meter(*args, **kwargs)
         len_s = len(s)
         fp.write('\r' + s + ' ' * max(last_printed_len[0] - len_s, 0))
         fp.flush()
@@ -333,9 +337,9 @@ class tqdm(object):
             # Initialize the screen printer
             self.sp = StatusPrinter(self.fp)
             if not disable:
-                self.sp(format_meter(0, total, 0,
+                self.sp(0, total, 0,
                         (dynamic_ncols(file) if dynamic_ncols else ncols),
-                        self.desc, ascii, unit, unit_scale))
+                        self.desc, ascii, unit, unit_scale)
 
         # Init the time/iterations counters
         self.start_t = self.last_print_t = time()
@@ -405,11 +409,10 @@ class tqdm(object):
                                 else smoothing * delta_it / delta_t + \
                                 (1 - smoothing) * avg_rate
 
-                        sp(format_meter(
-                            n, self.total, elapsed,
+                        sp(n, self.total, elapsed,
                             (dynamic_ncols(self.fp) if dynamic_ncols
                              else ncols),
-                            self.desc, ascii, unit, unit_scale, avg_rate))
+                            self.desc, ascii, unit, unit_scale, avg_rate)
 
                         # If no `miniters` was specified, adjust automatically
                         # to the maximum iteration rate seen so far.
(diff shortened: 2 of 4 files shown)


lrq3000:
Note that https://github.com/z-m-k/tqdm/commit/637156169c2b861a3bcfd73bf9e1e6a29202930b wasn't used here, I implemented it but it simply shown a percentage indicator, no progress bar, and only in plain text (so it was in fact less useful than just using `tqdm(ascii=True)`...).

I inspired my work by what whas done by flying-sheep:
https://github.com/flying-sheep/smart-progress

But I mainly just looked the docs and the sourcecode of the [ipywidgets repo](https://github.com/ipython/ipywidgets), where the Jupyter widgets are now separately developped out of Jupyter core codebase.

lrq3000:
Worth of note, there is also the ipy-progressbar module, which supports IPython as well as terminal, and it seems it supports **recursive progress bars** (which may be a useful feature if you ask me! - but I didn't try it at all): https://github.com/aplavin/ipy-progressbar

And two interesting SO posts to make our own widgets dispositions using standard widgets (which I had to do, there was no way to place the description text of the progress bar to the right...):

- http://stackoverflow.com/questions/30398873/how-to-left-align-widget-label-in-ipython-3
- http://stackoverflow.com/questions/26534123/aligning-textbox-widgets-in-ipython-notebooks

And of course the original IPython dev's post on how to make static widgets using HTML/JSS/CSS:
https://jakevdp.github.io/blog/2013/12/05/static-interactive-widgets/

lrq3000:
There's a bug when trying to use `tqdm_notebook()` and to manually update. I'll fix that tomorrow.

lrq3000:
Update about the last bug I meet: the issue is almost exactly described here:
https://thingspython.wordpress.com/2010/09/27/another-super-wrinkle-raising-typeerror/

I tested the code, and I run into the same bug:
```
>> t = tnrange(1)
<class 'tqdm._tqdm_notebook.tqdm_notebook'>
<class 'tqdm._tqdm_notebook.tqdm_notebook'>
True

>> t = tqdm_notebook()
<class 'tqdm._tqdm_notebook.tqdm_notebook'>
<class 'tqdm._tqdm_notebook.tqdm_notebook'>
False
```

But I don't know why it happens. In the blog post above, they say their issue was because of reimporting the same module multiple times, but here it does not seem to be the case...

lrq3000:
Fixed (without needing any commit): in fact, it was I activated IPython autoreloading of modules, which triggered a reimport of the module. What is weird is that it apparently did a partial reimport: tnrange() was updated, but not tqdm_notebook, which caused the issue.

So in the end, I just had to restart my IPython kernel, and all works now. Please note that the following also fixed the bug:

```
super(self.__class__, self).__init__(*args, **kwargs)
```

Instead of:

```
super(tqdm_notebook, self).__init__(*args, **kwargs)
```

But this is not a good fix as it can produce infinite loops, as demonstrated by this code snippet:

```
class A(object):
    def foo(self):
        print("In A")

class B(A):
    def foo(self):
        print("In B")
        super(self.__class__, self).foo()

class C(B):
    def foo(self):
        print("In C")
        super(self.__class__, self).foo()

c = C()
c.foo()
```

Also, the solution offered by the blog post above did not fix the issue for us:
```
self.as_super = super(tqdm_notebook, self)
self.as_super.__init__(*args, **kwargs)
```

lrq3000:
WOW, I just tested the [ipy-progressbar](https://github.com/aplavin/ipy-progressbar) module, and it's very impressive! Check out their ipython bars:

tqdm-ipython2

Our bar works, but maybe we can try to make it prettier like theirs? ;) Anyone willing to give it a try?

lrq3000:
I tested the performance, and it seems that normal tqdm isn't impacted. Please @casperdcl check that I'm not mistaken :)

About the performance of `tqdm_notebook`, it seems, surprisingly, that the performance drop isn't really significative (at first it can be ~ half the normal speed, but after a few iterations it converges to about the same speed as tqdm).

The code I used to benchmark:

```
from tqdm import trange, tnrange
from StringIO import StringIO
import sys
class MockFileNoWrite(StringIO):
    """Wraps StringIO to mock a file with no I/O"""
    def write(self, data):
        return
%timeit for i in trange(40000, desc='Bar desc', file=StringIO(), miniters=1, mininterval=0, leave=True): pass
%timeit for i in trange(40000, desc='Bar desc', file=MockFileNoWrite(), miniters=1, mininterval=0, leave=True): pass
for i in trange(10000, desc='Bar desc', file=sys.stdout, miniters=1, mininterval=0, leave=True): pass
for i in tnrange(10000, desc='Bar desc', file=sys.stdout, miniters=1, mininterval=0, leave=True): pass
```

lrq3000:
I implemented ipy_progressbar pretty progress bars inside `tqdm_notebook_pretty()` and `tnprange()`.  The pretty version uses a custom HTML/JS/CSS template, and for the moment I cannot find how to delete previous bars, so the following code will create more and more bars (just like ipy_progressbar in fact), but I think we can get around this issue:

```
from tqdm import tnprange
from time import sleep
for i in tnprange(10, desc='1st loop'):
    for j in tnprange(100, desc='2nd loop'):
        sleep(0.01)
```

Please try them and tell me which one we should keep between `tqdm_notebook` and `tqdm_notebook_pretty`.

lrq3000:
I need your feedback guys :) What version should we keep (native ipython widget or custom beautiful widget)? And also has anyone any idea how we could unit test that, or should we just skip from unit test for the moment?

casperdcl:
I'm not a notebook person so can't help out here, sorry

aplavin:
I am the author of ipy-progressbar, and I think it's better to use ipywidgets now. They have almost all capabilities needed, and lead to simpler and faster code. Why simpler is clear, and why faster you may ask? Because they don't require displaying javascript for each update, using sockets for it. I actually updated my progressbar today to use ipywidgets, and the only feature missing is that it's impossible to have two stacked progressbars (like at the image in this thread - the first bar shows the progress, and the second shows the current step). So, as tqdm seems to be really all about speed :) I think ipywidgets is the way to go.
I used my progressbar for scientific computations mainly, and they tend to take many orders of magnitude time more than any progressbar implementation. Btw, a feature I missed in other progressbar is the ability to show a custom (short) string at the progressbar - nice to see tqdm implemented it too :) Also, as I just came across tqdm - can it show speed in `sec per iteration` instead of `iterations per sec`?

casperdcl:
@aplavin thanks for your helpful comments. We automatically switch between it/s and s/it depending on which is > 1.

lrq3000:
@aplavin Hello and thank you very much for your very helpful comments :) Indeed I suspected ipywidgets would be faster by using some kind of direct communication via sockets, but I couldn't find any info on that.

Ok so that settles the matter. I'm going to sync this branch with master, and use the new features (eg, `nobar` will be replaced by `bar_format`).

lrq3000:
Continued in #92.

lrq3000:
Please do not delete this branch, this may be a useful starting point to create a web-friendly `tqdm` (indeed, the pretty notebook is really simply CSS+JSS, so we can probably reuse that to make a web-friendly printing for `tqdm`).

lrq3000:
Arg, I already deleted the interesting patch, but I have a local copy :) I'm going to open a new PR.

PR #63, opened 2015-12-01
Nested progressbars

### LINKED ISSUE #62: Support for nested progress bars + readme updates
tqdm should support displaying nested progress bars, in particular when using nested loops:

```
from tqdm import trange
from time import sleep
for i in trange(10, desc='outer loop', leave=True):
    for j in trange(100, desc='inner loop'):
        sleep(0.01)
```

This currently does not work because the outer loop is always deleted and replaced by the inner loop.

A simple fix would be to allow StatusPrinter to avoid using the `\r` replacing character when displaying the inner loop to not replace the outer loop. This is the approach taken by [ipy-progressbar](https://github.com/aplavin/ipy-progressbar).

Another way would be to update both bars independently, which is currently done for tqdm_notebook. But I'm not sure this is possible inside a terminal (is there a way to replace not only the current line but also the x previous lines? ie, the opposite of a line return).

We should then update the README to add this use case.

Also, we should add in the README that tqdm does not require any dependency, in particular curses (which is commonly required for various console utilities), this would fix https://github.com/noamraph/tqdm/issues/16.

lrq3000:
Fix #62 and https://github.com/noamraph/tqdm/issues/16.

This allows to do this:

tqdm-nested1

And even this:

tqdm-nested2

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -293,8 +293,25 @@ Here's an example:
     from time import sleep
 
     for i in trange(10, desc='1st loop', leave=True):
-        for j in trange(100, desc='2nd loop', nested=True):
+        for j in trange(100, desc='2nd loop', leave=True, nested=True):
             sleep(0.01)
+
+You can of course use as many nested loops as you want, get funky:
+
+.. code:: python
+
+    from tqdm import trange
+    from time import sleep
+
+    for i in trange(10, desc='1st loop', leave=True):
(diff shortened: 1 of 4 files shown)


lrq3000:
This patch is completed. Can someone please try it on Linux/MacOSX and report if it works alright? :)

PR #68, opened 2015-12-20
fix nested + leave completed cursor placement

lrq3000:
Example usage:

```
from tqdm import trange
from time import sleep
lv = True
ns = True
for i in trange(2, desc='outer0 loop', leave=True, nested=4):
    for j in trange(3, desc='inner1 loop', leave=lv, nested=ns):
        for k in trange(4, desc='inner2 loop', leave=lv, nested=ns):
            for l in trange(100, desc='inner3 loop', leave=lv, nested=ns):
                sleep(0.005)
```

Need to:
- [x] Fix unit test for 100% coverage
- [ ] rebase and fix conflicts

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -198,10 +198,10 @@ Parameters
     Exponential moving average smoothing factor for speed estimates
     (ignored in GUI mode). Ranges from 0 (average speed) to 1
     (current/instantaneous speed) [default: 0.3].
-* nested  : bool, optional  
+* nested  : bool/int, optional  
     Whether this iterable is nested in another one also managed by
     `tqdm` [default: False]. Allows display of multiple, nested
-    progress bars.
+    progress bars. Most outer loop can specify total nesting depth.
 
 Returns
 ~~~~~~~
@@ -352,14 +352,15 @@ Nested progress bars
 
 ``tqdm`` supports nested progress bars, you just need to specify the
 `nested=True` argument for all tqdm instantiations except the **outermost**
-bar. Here's an example:
+bar, which should specify `nested=x` where x is the total number of nested
+loops to correctly place the cursor after the bars finish. Here's an example:
 
 .. code:: python
 
     from tqdm import trange
     from time import sleep
 
-    for i in trange(10, desc='1st loop', leave=True):
+    for i in trange(10, desc='1st loop', leave=True, nested=3):
         for j in trange(5, desc='2nd loop', leave=True, nested=True):
             for k in trange(100, desc='3nd loop', leave=True, nested=True):
                 sleep(0.01)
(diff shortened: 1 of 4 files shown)


casperdcl:
hmm I like the simplicity of this fix. better than the dispatcher, I think.

lrq3000:
Ok thank's @caspercl :) I will fix the coverage tomorrow.

2015-12-20 18:28 GMT+01:00 Casper da Costa-Luis <notifications@github.com>:

lrq3000:
Fixed unit test and updated documentation. Ready to merge :)

lrq3000:
@casperdcl could you please tell me how you amend to add "bump version" message and then push to the master without creating a new branch to do that? Everytime I try, it rejects me because the master branch is protected now (which is good).

casperdcl:
Oh @lrq3000 you should be able to first merge master into this branch, fix conflicts, then push to this branch. Once all checks pass you should be able to merge the PR into master from here

casperdcl:
I'll temporarily unprotect master and rebase the ugly commit tree since the last release once this PR is fixed

lrq3000:
Ok @casperdcl , I thought I should never merge a PR from github's web interface, but I understand that if I do the merge manually, there won't be any ugly merge message aftewards :)

About the `leave=False` issue, I couldn't find it, it seems everything is running smoothly for me. I'm going to add a unit test for that. Could you please give me a code snippet to reproduce your issue? Here's mine:

```
from tqdm import trange
from time import sleep
lv = False
ns = True
for i in trange(2, desc='outer0 loop', leave=True, nested=4):
    for j in trange(3, desc='inner1 loop', leave=lv, nested=ns):
        for k in trange(4, desc='inner2 loop', leave=lv, nested=ns):
            for l in trange(100, desc='inner3 loop', leave=True, nested=ns):
                sleep(0.005)
```

Result:
```
outer0 loop: 100%|###############################| 2/2 [00:12<00:00,  0.16it/s]

inner3 loop: 100%|##########################| 100/100 [00:00<00:00, 198.50it/s]
C:\git\rfigc\tqdm>
```

lrq3000:
The latest commit should fix all issues with leave and nested.

casperdcl:
well I was referring to the blank lines (2 between outer0 and outer3 in your example). Would be nice to somehow delete them.

lrq3000:
Ah I see, well, I don't think that's possible to do with our decentralized approach to nested printing, each bar has no knowledge of how many other bars there are out there, so it can just print on its own line and give the hand to the parent bar, that's why they are left in place. Personally, I prefer that the bars stay at the same place after leave, so this issue doesn't bother me :)

lrq3000:
To fix this issue and #87, we would need to use a centralized approach, the dispatcher concept. This would allow us to have total positionning control, just like in a curses environment. Of course, this flexibility will cost in terms of performance.

lrq3000:
Closing this PR, see new approach in #87. We should delete this branch after the other PR gets merged.

PR #90, opened 2015-12-29
Add performance test via bytecode opcode analysis

lrq3000:
This PR provides dynamic bytecode opcodes analysis in `tests_perf.py`, to compare the number of opcodes required to run `tqdm` vs a native `range()`.

The idea is to run `tqdm` and dynamically analyze the number of bytecode opcodes executed at runtime for a standard run. This would provide a way to discretize and virtualize the time taken to run `tqdm` compared to native `range`.

The problem is that it works only on Python 2.x because the required library [bytecode_tracer](https://github.com/mkwiatkowski/bytecode-hack) requires `implib`, but `implib` was removed from Python 3.x and replaced by `importlib`, but the API is quite different. On the other hand, `importlib` wasn't backported to Python 2.x. We should thus rewrite `code_rewriting_importer.py` to use `implib` on Py2 and `importlib` on Py3, I don't think it should be a lot of work for someone with experience in making `custom importers`, but I don't have this experience so I cannot do much more.

TODO:
- [ ] Fix [bytecode_tracer](https://github.com/mkwiatkowski/bytecode-hack) Py3 compatibility (ie, need to use `importlib` instead of `implib` in `code_rewriting_importer.py`).
- [ ] Fix nosetests trying to test `bytecode_tracer` module.

### DIFF
--- a/tqdm/tests/bytecode_tracer/__init__.py
+++ b/tqdm/tests/bytecode_tracer/__init__.py
@@ -1,0 +1,3 @@
+from .bytecode_tracer import BytecodeTracer, rewrite_function
+
+__all__ = ['BytecodeTracer', 'rewrite_function']
--- a/tqdm/tests/bytecode_tracer/bytecode_tracer.py
+++ b/tqdm/tests/bytecode_tracer/bytecode_tracer.py
@@ -1,0 +1,293 @@
+import opcode
+import os
+import re
+
+from types import CodeType, MethodType
+
+from . import code_rewriting_importer
+
+from .py_frame_object import get_value_stack_top
+
+
+class ValueStack(object):
+    """CPython stack that holds values used and generated during computation.
+
+    Right before a function call value stack looks like this:
+
+        +---------------------  <--- frame.f_valuestack
+        | function object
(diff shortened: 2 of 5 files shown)


lrq3000:
Interesting links:
- https://docs.python.org/3/library/importlib.html
- http://bioportal.weizmann.ac.il/course/python/PyMOTW/PyMOTW/docs/sys/imports.html
- http://blog.dowski.com/2008/07/31/customizing-the-python-import-system/
- http://stackoverflow.com/questions/27307072/what-is-the-point-of-pythons-importlib
- https://pymotw.com/2/sys/imports.html
- http://fr.slideshare.net/Zoom.Quiet/import-this-that-and-the-other-thing-custom-importers
- http://sayspy.blogspot.fr/2009/06/lazy-imports-in-25-lines-of-code.html

lrq3000:
Another library that can maybe be used instead: [glassbox](https://github.com/DRMacIver/glassbox).

lrq3000:
A new library: [py.code](http://pylib.readthedocs.io/en/latest/code.html)

PR #91, opened 2015-12-29
Async coroutine support

lrq3000:
Add asynchronized coroutine support in `tqdm` as proposed in #65.

Code thank's to @aplavin, from [here](https://github.com/aplavin/ipy-progressbar/blob/5c2f9a8cfe1e92e0ec2ceb741e92519e73d5b6b7/ipy_progressbar/base.py) and [here](https://github.com/aplavin/ipy-progressbar/commit/324148943f3a133b1d34325f36302673a6d10649).

This code wasn't tested out, can someone devise an example case? (I'm inexperienced with Python coroutines :S).

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -12,7 +12,7 @@ from __future__ import absolute_import, division
 from .utils import _supports_unicode, _screen_shape_wrapper, _range, _unich, \
     _term_move_up, _unicode, WeakSet, _basestring, _OrderedDict, \
     Comparable, _is_ascii, FormatReplace, disp_len, disp_trim, \
-    SimpleTextIOWrapper, CallbackIOWrapper, _coroutine
+    SimpleTextIOWrapper, CallbackIOWrapper, coroutine
 from ._monitor import TMonitor
 # native libraries
 from contextlib import contextmanager
@@ -1095,12 +1095,12 @@ class tqdm(Comparable):
     def __hash__(self):
         return id(self)
 
-    @_coroutine
+    @coroutine
     def __aiter__(self):
         self.start()
         return self
 
-    @_coroutine
+    @coroutine
     def __anext__(self):
         try:
             self.update()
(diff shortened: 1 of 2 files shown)


sametmax:
import asyncio

    from tqdm import tqdm

    async def main():
        async for x in tqdm(range(10)):
            print(x)

    loop = asyncio.get_event_loop()
    loo.run_until_complete()

How ever, there is much more to consider to unit test it:

- __aiter__ is only awailable on 3.5 and requires async/await, which are SyntaxErrors on earlier versions. So the test must be in a separate files, and ignored by the test runner for python 3.4-.
- __anext__ should probably do "yield from self.iterable" instead of next() in case the iterable itself is asyncronous.
- if you want to unit test it, you'll need to start the eventloop as a setUp and stop it as a tearDown. Libs like pytest-asyncio do it automatically for you.
- if you need to mock an awaitable, do:

    import asyncio

    class AMock(Mock):

        async def __call__(self, *args, **kwargs):
            result = super().__call__(self, *args, **kwargs)
            async def coroutine():
                return result
            return asyncio.ensure_future(coro())

lrq3000:
Wow thank you very much sametmax, this sure helps a lot ! I'll train a bit
with toy examples following your guidelines before making a unit test, and
ill change next() for the more pythonic syntax in my PR :)

AlJohri:
Just ran into this issue myself. Bumping this PR.

```
TypeError: 'async for' requires an object with __aiter__ method, got tqdm
```

tqdm/_tqdm.py
```diff
@@ -409,6 +409,20 @@ def __exit__(self, *exc):
         self.close()
         return False
 
+    @_coroutine
```
AlJohri: I believe this one should not be a coroutine.

AlJohri:
This works for me:

```python
from tqdm.auto import tqdm

def __aiter__(self):
    return self

async def __anext__(self):
    try:
        self.update()
        return await self.iterable.__anext__()
    except StopIteration:
        self.close()
        raise StopAsyncIteration
tqdm.__aiter__ = __aiter__
tqdm.__anext__ = __anext__

################################################

async for row in tqdm(con.cursor('SELECT * FROM pageviews ORDER BY user_id')):
    pass
```

casperdcl:
ah nice, I suppose I should take over. @AlJohri could you write a unit test?

malarinv:
@casperdcl will this be merged soon?

PR #92, opened 2015-12-29
Add ipython/jupyter notebook support (rebased)

lrq3000:
Replaces PR #61. Rebased and updated approach to use `bar_format` instead of modularizing `StatusPrinter` in `_tqdm.py` (so there's no modification of the core here). Maybe it will be a bit slower than previous PR (because of the additional string processing and because of using `bar_format`), but from my own tests it seems close to the speed of base `tqdm`.

Example usage (in a notebook):

Iteration-based:
```
from tqdm import tnrange
from time import sleep
for i in tnrange(10, desc='1st loop', leave=True):
    for j in tnrange(100, desc='2nd loop', leave=False):
        sleep(0.01)
```

Manual control:
```
from tqdm import tqdm_notebook
from time import sleep

with tqdm_notebook(total=10, desc='1st loop', leave=True) as bar1:
    for i in xrange(10):
        with tqdm_notebook(total=1000, desc='2nd loop') as bar2:
            bar2.set_description('2nd loop(%i)' % i)
            for j in xrange(100):
                sleep(0.01)
                bar2.update(10)
        bar1.update()
```

Todo:
- [x] Add to readme.
- [x] Unit test??? Nope, too much complicated, we would have to [mock an IPython JSON server as demonstrated by the ipywidgets devs](https://github.com/ipython/ipywidgets/blob/master/ipywidgets/test/widget_test.js).
- [x] Add `info` bar_style (reuse output from standard tqdm managing iterable without total)
- [x] Test bar styles


casperdcl:
@lrq3000 I think you still wanted to do other things here judging by the todo list...

lrq3000:
@casperdcl I wanted to unit test but I really have no idea how to do it, so maybe we can leave this out for later. I'll add to the README and I'll leave you with bumping the version.

lrq3000:
Ok I'm done with this patch. I would prefer to have a unit test, but it's way too much complicated and this would add two dependencies (ipywidgets and ipython), so I guess it's better to leave that out of the picture for the moment (I'd rather focus on getting tqdm bugfree).

casperdcl:
Yes, this is the same reason I didn't include a gui test.

alexanderkuk:
Here are some examples of how progress bar can look like in IPython Notebook https://github.com/alexanderkuk/log-progress#examples . Notice how it changes color based on outcome. Just thought it can be somehow useful for you

lrq3000:
Very nice indeed, I didn't know about bar_style, thank's a lot, this will
be added in this PR!

2016-02-08 23:01 GMT+01:00 Alexander Kuk <notifications@github.com>:

lrq3000:
Thank you a lot @alexanderkuk, I implemented all the bar styles, the module is a lot better now (I added you as co-author).

This PR is finished. When #150 gets merged, I can simplify one line of the code (thank's to `self.refresh()`), but the module will stay the same. Can be merged now or after #150.

lrq3000:
Ok I'm all done, this PR is ready to merge! I hope IPython users will be pleased :)

lrq3000:
I added screenshots to demonstrate the Jupyter submodule.

If noone argues, I will merge this PR on next week-end.

lrq3000:
Should I merge this and the other PRs?

casperdcl:
srry @lrq3000 never got around to reviewing this. thanks for the work!

lrq3000:
np @casper, I use IPython quite often anyway so that was easier for me :)

2016-05-15 14:31 GMT+02:00 Casper da Costa-Luis <notifications@github.com>:

casperdcl:
@lrq3000 I get this warning ever time `tqdm` is used now (since ipython utils are imported in `__init__.py`):

```
/usr/local/lib/python2.7/dist-packages/IPython/html.py:14: ShimWarning: The `IPython.html` package has been deprecated. You should import from `notebook` instead. `IPython.html.widgets` has moved to `ipywidgets`.
  "`IPython.html.widgets` has moved to `ipywidgets`.", ShimWarning)
```

lrq3000:
Damn, we are suffering the dependency hell. Since we are likely to add more
and more modules, maybe we should find a definite solution for this issue
that will only grow in the future? I'm fixing it later today.

2016-05-15 15:35 GMT+02:00 Casper da Costa-Luis <notifications@github.com>:

casperdcl:
yes this is why it certainly wasn't np that I didn't get to review this. definitely can't push this to pypi before fixing.

lrq3000:
Ok thank you very much Casper, I pushed your fix, indeed that was very
weird because Travis didn't say anything about this warning (it doesn't
have IPython as a dependency so it should have happened). But I think we
really ought to find a way to avoid loading unrequired submodules when just
importing core `tqdm`. I'll look into that later.

2016-05-15 15:47 GMT+02:00 Casper da Costa-Luis <notifications@github.com>:

PR #94, opened 2015-12-29
Web-friendly tqdm

lrq3000:
Implements a web-friendly tqdm by using JSS to dynamically redefine JSS. Based on aplavin's great work on [ipy-progressbar](https://github.com/aplavin/ipy-progressbar).

The output is correct, but I don't know how we can test that out (ie, show this in a web page). Someone has an idea?
- [ ] Find a way to test that in an html page.
- [ ] Unit test.
- [ ] Add to readme

### DIFF
--- a/tqdm/_tqdm_web.py
+++ b/tqdm/_tqdm_web.py
@@ -1,0 +1,155 @@
+"""
+Web-friendly progressbar decorator for iterators.
+Includes a default (x)range iterator printing to stderr.
+
+Usage:
+  >>> from tqdm_web import twrange[, tqdm_web]
+  >>> for i in twrange(10): #same as: for i in tqdm_web(xrange(10))
+  ...     ...
+"""
+# future division is important to divide integers and get as
+# a result precise floating numbers (instead of truncated int)
+from __future__ import division, absolute_import
+# import compatibility functions and utilities
+import sys
+import uuid
+from ._utils import _range
+
+# HTML encoding
+try:  # pragma: no cover
+    from html import escape  # python 3.x
+except ImportError:  # pragma: no cover
+    from cgi import escape  # python 2.x
+
+# to inherit from the tqdm class
+from ._tqdm import tqdm, format_meter, StatusPrinter
+
+
+__author__ = {"github.com/": ["lrq3000", "aplavin"]}
+__all__ = ['tqdm_notebook', 'tnrange']
+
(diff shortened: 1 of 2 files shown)


tqdm/_tqdm_web.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class tqdm_web(tqdm):  # pragma: no cover
+    """
+    Experimental CSS/JSS web-friendly tqdm!
+    """
+    def __init__(self, *args, **kwargs):
+
+        # Setup default output
+        if not kwargs.get('file', None) or kwargs['file'] == sys.stderr:
```
o11c: `kwargs.get('file', sys.stderr) is sys.stderr`

casperdcl: thanks. looks like this was copied from the current notebook implementation.

lrq3000: Actually it was extracted from a variant of _tqdm_notebook. Whereas the
current _tqdm_notebook leverages ipywidgets to show and update the progress
bars, this variant defined its own JS+CSS implementation. However, this
implementation was way slower than ipywidgets, hence why I reverted back,
but this could be used for a web implementation of tqdm (the code is
already fully functional, I just need some feedback from a python web dev
on how to implement the API).

2016-05-18 8:55 GMT+02:00 Casper da Costa-Luis <notifications@github.com>:

CrazyPython:
@lrq3000 We'll need an intermediary server to communicate between the two. You could use one of the several server frameworks for python (flask, tornado, django, etc.)

lrq3000:
@CrazyPython Yes we need a server framework to test the bar, but I don't intend the bar to provide the full stack to show the bars in a web page, but rather just the HTML/CSS/JS output that the user can then plug in any server framework of choice. This would limit the maintenance cost for us and also be more generic. If that's possible...

CrazyPython:
@lrq3000 I was imagining a simple server on `localhost`

lrq3000:
Yes of course it's necessary to be localhost so that we can run unit tests,
but I mean that we should use such a server framework only for unit tests,
but the web-friendly tqdm submodule should be server-agnostic so that it
can work everywhere! (in an ideal world haha...).

2016-08-05 0:37 GMT+02:00 CrazyPython <notifications@github.com>:

PR #102, opened 2016-01-05
add all python versions to .travis.yml

PhillipSz:
From: https://docs.travis-ci.com/user/languages/python#Choosing-Python-versions-to-test-against

If they are all supported why not test them?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,14 @@
 sudo: false
 language: python
-python: 3.5
+python:
+  - "2.6"
+  - "2.7"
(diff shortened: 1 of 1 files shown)


lrq3000:
The different python versions are declared in tox, so in fact in travis only the base python for the virtualenv is defined. I don't know if adding more base python versions in travis change something from our current setup. Do you have more information about this?

PhillipSz:
right, no that should be fine. Sorry for for my fast useless PR.

Anyway thanks for tqdm!

PR #112, opened 2016-01-21
Fixed setup.py so that it doesn't throw encoding error.

shirish93:
Without this change, the installation via python setup.py install is broken in certain Windows system with encoding set to other than 'utf-8'.

Without specifying the encoding of the README file, the open defaults to system config encoding, which doesn't have to be 'utf-8'. Opening the file in UTF-8 mode fixes that.

Traceback of the error thrown without the fix:

`Traceback (most recent call last):
  File "setup.py", line 156, in <module>
    README_rst = fd.read()
  File "C:\Users\n0272436\Downloads\WinPython-32bit-3.4.3.4\python-3.4.3\lib\encodings\cp1252.py", line 23,
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8d in position 17233: character maps to <undefined>
`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -152,7 +152,8 @@ if sys.argv[1].lower().strip() == 'make':
 
 
 README_rst = ''
-with open('README.rst', 'r', encoding = 'utf-8') as fd:
+import io
+with io.open('README.rst', mode = 'r', encoding = 'utf-8') as fd:
     README_rst = fd.read()
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -152,7 +152,7 @@ if sys.argv[1].lower().strip() == 'make':
 
 
 README_rst = ''
-with open('README.rst', 'r') as fd:
+with open('README.rst', 'r', encoding = 'utf-8') as fd:
     README_rst = fd.read()
 
 setup(


shirish93:
So it appears that the build fails due to this pip error:
`
Support for Python 3.0-3.2 has been dropped. Future versions will fail here.
`
I see this has been discussed here: https://github.com/john-kurkowski/tldextract/issues/50

Perhaps opening the file with UTF-8 encoding fails in 3.0-3.2? Not sure why my code would cause this exact behavior. 

Either way, the failure of test 566.3 seems to pit all other versions of Python against 3.0-3.2. Would it be worth considering to remove those from build tests entirely?

lrq3000:
Thank you for the patch, i'm the dev running on windows so ill try your
patch when i get internet access with my computer.

About dropping Python 3.2, yes we will do but only when a new PyPy3 will be
released. Until then,  i'd like to support both.
Le 21 Jan. 2016 18:14, "Shirish Pokharel" <notifications@github.com> a
écrit :

lrq3000:
Ps: moderate my previous reply : we will MAYBE drop Py3.2, when a new PyPy3
gets released. We will considerate that from the active userbase (but I
know Py3.2 is widely seend as an incomplete version so yeah,  if noone uses
it, we'll drop).

casperdcl:
Travis is failing to build the environment (see #117 ) so I'm going to merge this in. master is currently failing due to this issue anyway.

lrq3000:
Ok thank's for confirming the issue and the culprit. I agree, let's drop Py32. It's way too buggy anyway, even in normal circumstances.

lrq3000:
Thank you @shirish93 for the patch!

lrq3000:
What's interesting is that PyPy3, which is based on Py32, doesn't suffer from this issue. So it's all ok for me.

PR #143, opened 2016-04-09
Multi tqdm

cscanlin:
This is a handler and interface for running multiple progress bars simultaneously. It iterates through progress bars and provides useful hooks that can be used to update and return results dynamically.

I'm really not sure if this is the best way to go about things. The whole wrapper/interface idea still feels a little clunky. There is probably a better way to build all of this functionality into tqdm properly. Definitely open to feedback on better ways to do any/all of this.

Issues:
- Need to test if it can handle being passed iterables
- When a Bar calls `close()`, it messes up other bars. (This is somewhat fixed with the clear and refresh)
- Need to handle moving command line cursor to below bars
- Needs compatibility testing

Considerations:
- this would work well with the print-message branch. The clear and refresh methods are from there as well.
  - How to handle async
  - Naming conventions?
  - Better way to handle task registry

Here is a minimum example (see tests/tests_multi.py for a more detailed example):

```
from tqdm import multi_tqdm, tqdm_job

class MiniTestJob(tqdm_job):

    def __init__(self, task_num, **kwargs):
        super(MiniTestJob, self).__init__(**kwargs)
        self.task_num = task_num

    def update(self):
        self.pbar.update(n=self.task_num)

def mini_test():
    multi = multi_tqdm()
    for task_num in range(1, 10):
        job = MiniTestJob(task_num=task_num, desc=str(task_num), total=1000)
        multi.register_job(job)
    multi.run(sleep_delay=.001)
```


cscanlin:
I've come to realize that this is really two separate features: tqdm_multi, and the callback features. tqdm_multi definitely supports these callbacks, but I think the callbacks are much more interesting and could be used outside of the tqdm_multi context. Essentially, a user can subclass tqdm and define logic around what to do when a progress bar finishes, and can handle the results or errors as they happen. This is especially useful in async operations, so more support around that could prove very useful.

One issue is that I'm not sure exactly where to put the `run_callbacks()` function when not running under tqdm_multi, so any advice there would be appreciated. I'm also unsure about the best way to figure out if a tqdm instance is still running (right now I'm using `tqdm.n < tqdm.total` but I think there must be a better way)

casperdcl:
There used to be `tqdm.closed` but this was found to be redundant since
whenever it was checked it's value was equal to `tqdm.disable`... which you
could use instead.

async usage would be great; I haven't had time to look into it yet.

lrq3000:
About async, shouldn't the effort be merged with #91 ? I feel like this PR is trying to implement many features at once, maybe break it up into separate, more focused PRs would make things easier (and cleaner)?

lrq3000:
To precise my idea: async should be a native feature of tqdm anyway, so I expect that implementing it in tqdm would make things more elegant and your life easier at the same time.

But great job for tqdm_multi, it's quite clean as it is, and nicely separated onto its own module.

AkihiroSuda:
any update on this?

cscanlin:
Hey, I know it's been awhile, but it would be awesome if we could re-examine this.

I think the major changes to be made are:

1. Refactor `_is_complete` to be a better match for what we're checking for.

2. Remove references to `tqdm_job` since that portion is mostly abandoned. It looks like this logic has been implemented with [hooks and callbacks](https://github.com/tqdm/tqdm#hooks-and-callbacks) now, so there may be ways to integrate that now, or at a later date.

3. Add an example to the README

I'm making this comment in reference to a post on the /r/learnpython that seems to be asking for almost exactly this use case, so it appears there is still demand for it: https://www.reddit.com/r/learnpython/comments/5l7rpn/trying_to_wrap_my_head_around_a_problem_just/

Anyways, just wanted some feedback before I dig back into this, let me know if anyone has any thoughts!

lrq3000:
@cscanlin Thank you for coming back to this PR :)

I think I better understand what your PR aims to do: a centralized manager for progress bars display running in parallel with the code. If this is what you intended, then I understand the difficulties: the philosophy is at the opposite of `tqdm`'s, because `tqdm` aims to be decentralized (ie, there is almost no knowledge of other instances, each bar management is independent from the others, and the progress bar display is tightly integrated in for loops so it cannot be put in a centralized manager). All other progress bars use the centralized philosophy, but we did not because the decentralized philosophy is much more efficient (because you do not need to call `time.time()` for most of the calls, thanks to `miniters` trick, which cannot be done with a centralized approach).

Having said that, your PR is very elegant, and is separated into another module, so why not provide this as an additional, optional feature? :)

But identifying the philosophy allows to better identify the goal and the issues:
* You need to know when bars complete to remove them from the pool of jobs of the centralized manager. The current approach (checking if n >= total) is not reliable. You should indeed reinstate as Casper suggested a `tqdm.completed = True` flag that would be set inside `tqdm.close()`.
* Indeed `tqdm.write()` could be implemented, quite easily in fact: just add a method like this is `tqdm_multi`:
```
clas tqdm_multi(object):
# ...
    def write(*args, **kwargs):
        tqdm.write(*args, **kwargs)
```

`tqdm.write()` is a class method, so you don't need to instanciate any tqdm bar to be able to call it (it will obviously don't print anything but I mean that you don't have to manage anything from `tqdm_multi`).

* Use parallelism lock (see #329 ): since the manager is centralized, you can take advantage of this to provide a write lock automatically for parallelized applications as seems to be the case in the [reddit post you linked](https://www.reddit.com/r/learnpython/comments/5l7rpn/trying_to_wrap_my_head_around_a_problem_just/).
* `tqdm_multi` should support any `tqdm` module, it seems so currently so that's good. The unit test however can be done only on core `tqdm`, that's good enough.

Just my feedback based on my insight, but if I'm wrong please let me know so I can better understand what you try to achieve :) Also the PR is in a very good track, please keep up your good work!

lrq3000:
In the end, I think this PR should be a very good and necessary complement to #329, if only because of `write()` that is not manageable with only a decentralized approach when we use multiprocesses.

tqdm/_tqdm_multi.py
```diff
@@ -0,0 +1,56 @@
+import time
+
+from ._tqdm import tqdm
+
+"""
+**NOTES**
```
CrazyPython: Comment should be improved.

tqdm/_tqdm.py
```diff
@@ -757,6 +760,9 @@ def set_description(self, desc=None):
     def moveto(self, n):
         self.fp.write(_unicode('\n' * n + _term_move_up() * -n))
 
+    def _is_complete(self):
+        """(Needs refactoring) A way to see if a progress bar and/or task has finished"""
```
CrazyPython: Why does this need refactoring? `TODO: needs refactoring` at the end may be better.

tqdm/_tqdm_multi.py
CrazyPython: Comment should be improved.

tqdm/_tqdm_multi.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+
+class tqdm_multi(object):
+    """
+    A handler for running multiple progress bars simultaneously.
+
+    See tests/tests_multi.py for an example
+    """
+    def __init__(self, jobs=[], **kwargs):
```
CrazyPython: Mutable default argument! This is a bug.

CrazyPython (changes requested):
:+1:  I suggest making these changes, though.

PR #153, opened 2016-04-23
Add callbacks to tqdm.close()

cscanlin:
See #143 

TODO:
- [ ] Re-add tests
- [ ] Add option to pass functions as params
- [ ] Figure out how to do with ascync
- [ ] Add readme examples


lrq3000:
Could you please describe quickly what you want to achieve with this PR, and provide if possible a minimal example?

cscanlin:
This is all pulled out in various forms from #143. I have a bunch of examples, descriptions and tests left over from that PR, I just need to get it organized still. Will update when I can.

cscanlin:
So the premise of this feature is to make it easier for the user to subclass tqdm and define logic around what to do when a progress bar finishes. The user can specify any processing that needs to be done after the progress bars finishes, and provide handlers for the results or errors as they happen. Here is a minimal example with some a few basic callbacks to handle variable input:

    from tqdm import tqdm
    import time
    import random

    class callback_tqdm(tqdm):

        def __init__(self, *args, **kwargs):
            super(callback_tqdm, self).__init__(*args, **kwargs)

        def handle_result(arg):
            rand = random.randint(1, 10)
            if rand < 5:
                raise NameError('{0} is Less than 5!'.format(rand))
            else:
                return '{0} is At Least 5!'.format(rand)

        def success_callback(self, result):
            print('Success!', result)

        def failure_callback(self, error):
            print('Failed!', error)

    for __ in range(5):
        for i in callback_tqdm(range(1, 100)):
            time.sleep(.01)

Where this really shines though is in combination with tqdm_multi in #143. So far most of my work has been with checking the progress of some long running tasks on a remote server, and then processing the results as they happen. With tqdm_multi it is possible to start a variety of tasks, update the progress bars sequentially and handle the results as they finish without interrupting the other tasks.
This could also fit in nicely with the official async functionality, but I'm not sure exactly how to go about that. Here is a more fleshed out example using tqdm_multi (requires merging in #143 as well):

    from tqdm import tqdm, tqdm_multi

    import random

    class TestJob(tqdm):

        def __init__(self, task_num, **kwargs):
            super(TestJob, self).__init__(**kwargs)
            self.task_num = task_num

        def update(self):
            super(TestJob, self).update(n=self.task_num)

        def handle_result(self):
            if self.task_num == 5:
                raise NameError('No 5s allowed')
            else:
                return 'Success: {self.task_num}\n'.format(self=self)

        def success_callback(self, result):
            self.multi.out.write(result)

        def failure_callback(self, error):
            self.multi.out.write('Failed {self.task_num} with error: "{error}"\n'.format(
                self=self, error=error)
            )
            self.multi.register_job(self.restart_job())

        def restart_job(self):
            new_task_num = self.task_num * 10 + 2
            return TestJob(task_num=new_task_num, file=self.fp, desc=str(new_task_num), total=1000)

    with open('results.txt', 'w') as out:
        multi = tqdm_multi(out=out)
        for task_num in range(1, 10):
            task_num = random.randint(1, 10)
            job = TestJob(task_num=task_num, desc=str(task_num), total=1000)
            multi.register_job(job)
        multi.run(sleep_delay=.001)

It still needs more work, but I'm interested in feedback, and please let me know if I can help clarify anything.

lrq3000:
Thank you @cscanlin for taking the time to provide examples and explanations, this is much clearer.

If I understand your intent, you'd like `tqdm` to manage errors/completion of a task, and provide users a way to make their own customized callbacks to handle these cases?

If that is correct, I'm not sure this is the purpose of a progress bar like `tqdm`. Wouldn't it be easier to just wrap `tqdm` into your own `TaskManager` class, with all the features you described above?

Also another issue that tickles me is that this feature would only work for iterable-based `tqdm`, not manually updated `tqdm`.

cscanlin:
I went through the same thought process when I was decoupling this functionality from `tqdm_multi`, but I think it provides value for a few reasons.

First of all if you look at the changes, there are only about half a dozen lines of actual code so the overhead is incredibly light.

What these changes really do is formalize and provide a minimal framework around callbacks so the user can implement this functionality and `tqdm` handles some of the basic boilerplate. This also makes it easier to help get people started with some examples of how this is done, and I think it's one of the coolest things `tqdm` can be used for (especially with `tqdm_multi`).

When I first started on this feature, I actually originally wrote it as a separate abstract base class (that inherited from `tqdm`), which the user would then inherit from. I realized after working on it a while that it's actually unnecessary to have the intermediate wrapper because the implementation works perfectly when inherited directly from `tqdm`.

Also, it will work with manual update, as long as they call `tqdm.close()` or `tqdm.run_callbacks()` after the bar is finished :) 

Let me know if I can clarify anything further!

lrq3000:
Ok thank you for the clarification. I agree It can be a very good addition
to tqdm as a submodule, but I'm still not sure it should make it in the
core tqdm :/ But you can continue your work on this PR as it is, we can
take care of turning it into a submodule.

Casper, do you have an opinion on this?

2016-05-05 0:10 GMT+02:00 Chris Scanlin <notifications@github.com>:

PR #154, opened 2016-04-23
Completion checking refactor

cscanlin:
Beginning of refactor to `tqdm._is_complete()` and `tqdm.update()` to separate the logic to check if is complete.

TODO:

- [ ] Add `_is_complete()` check to `__iter__`
- [ ] Remove extra variable assignments in `__iter__`?
- [ ] Add tests?


cscanlin:
This did not work as intended really, `_is_complete()` does not actually say whether it has finished, and it seems impossible to apply to `__iter__`. Closing for now

Issue #140, opened 2016-04-05
Option to show an exponential rate trend


graingert:
Exponentially increasing progress bars appear faster:

http://www.chrisharrison.net/projects/progressbars/ProgBarHarrison.pdf

it's probably a good idea to hide percentage calculations from these bars though.

casperdcl:
Thanks for this. Having skimmed through the paper I think we already do better. `tqdm` uses an exponential weighted moving average which minimises overhead but allows some flexibility. Use the `smoothing` keyword argument if you need more control!

graingert:
@casperdcl the smoothing docs aren't very clear on what it does.

lrq3000:
@graingert What you propose is not really the same as what is implemented: we implement an exponential moving average algorithm to better predict the time it will take to complete the task by better adapting to hiccups and changes in speeds, but your proposition is to make an ergonomically designed progress bar that is pessimistic so that it speeds up towards the end, giving the user the feeling that the process gets faster and faster.

We already discussed about this solution in the past, but we didn't implement it (yet?). I don't think we will implement it in the near future, but if you want to give it a try, I have found a [few posts that could give a good start here](https://github.com/tqdm/tqdm/issues/48#issuecomment-154848125).

lrq3000:
Actually, if we want to implement a pessimistic bar while still keeping the performance, I think we can make something out of just memorizing the slowest time and the fastest time per iteration, and we can just then apply this ratio on the time prediction given the number of iteration are left: the idea is that closer we are from the start, the more we will weight iterations that took a time closer to the slowest bound ; the closer from the end iteration, the more weight for iterations that are fast. This is pessimistic, and it requires only 2 new variables and 1 calculation.

casperdcl:
I'm personally against making progress appear "faster" - I'd much prefer
accuracy and precision and spreading the love of this. Imagine someone
using `tqdm` as a benchmarking utility - this won't go down well. (Sorry,
I'm a scientist... I also prefer SI units to barbaric ones.)

lrq3000:
@casperdcl agree, but the "ergonomic" feature could be an option we could disable. But yes, I agree that it's not a necessary one, I also prefer precision and benchmarking ability.

PR #158, opened 2016-04-26
Implementation of pessimistic progress rate

lrq3000:
I implemented a pessimistic progress rate, as asked in #140.

Performance-wise, with default settings, this just adds an additional `if` statement since it's disabled. When enabled, it's just a few more calculations and one more variable to store the worst time.

This implementation shows an exponential trend for the rate and remaining time, but not for the progress bar itself. For a full implementation, we would need to artificially avoid refreshing the bar, and refresh it more and more often towards the end (maybe by tweaking `min_iters`?).

TODO:
- [ ] Tweak `min_iters`?
- [ ] Add unit tests

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -314,6 +314,7 @@ class tqdm(object):
                  maxinterval=10.0, miniters=None, ascii=None, disable=False,
                  unit='it', unit_scale=False, dynamic_ncols=False,
                  smoothing=0.3, bar_format=None, initial=0, position=None,
+                 pessimistic=False,
                  gui=False, **kwargs):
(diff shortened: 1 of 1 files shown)


lrq3000:
I figured I didn't really describe what this PR does. So what it does is simply to store the worst iteration time ever, and it then calculates a mixture between the worst iteration time and the average iteration time (smoothed or not) based on the current progress: if we are close to the start, we will tend more to the worst iteration time ; if we are closer to the end, we will tend more to the average iteration time.

Thus, this doesn't show a fake time but rather a pessimistic time, hence the name (ie, the worst case time instead of the average time).

Issue #144, opened 2016-04-10
Provide an option about inv_rate


ppwwyyxx:
I notice that when rate is less than 1 unit /s, it will display as [number]s /unit.

Sometimes it may not be a good idea: I need to think about what unit it is to know if I'm getting faster or slower. I'd prefer a simple way like 0.7 unit / s.

Could there be an option for this? Thanks.

lrq3000:
We can add this option in custom bar formatting, would that be ok for you?

ppwwyyxx:
Sure! That would be great.

lrq3000:
Ok I implemented it in the [branch inv-option](https://github.com/tqdm/tqdm/tree/inv-option), please try it and tell me if this fits your need. Here is the sample code to generate the bar with not inversed rate:

```
from tqdm import trange
from time import sleep

for i in trange(10, bar_format="{l_bar}{bar}| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_noinv_fmt}]"):
    sleep(2)
```

ppwwyyxx:
It's really nice! That's what I needed.

lrq3000:
Ok then it will get merged :) Thank you for the feedback!

2016-05-09 17:09 GMT+02:00 Yuxin Wu <notifications@github.com>:

PR #159, opened 2016-05-08
Add no rate inversion option in bar_format

lrq3000:
Fix issue #144 by providing two variables `rate_noinv` and `rate_noinv_fmt` in `bar_format` to always display the "not inversed" rate.

Also, this PR fix the issue that the `rate` in `bar_format` was always non inversed, whereas it should be inversed when `rate < 1`.


lrq3000:
Confirmed by OP that it fixes the issue.

Issue #149, opened 2016-04-22
Bar formatting lacks flexibility leading to improper width


karanlyons:
Currently there are some optimizations in bar drawing that make the default fast, but these optimizations seem to impair formatting of custom bars, most noticeably (to me at least) in the width of the final string.

Instead of formatting the bar (by which I mean the actual progress bar and not any surrounding text) *first*, the bar should be formatted *last*, accounting for the format of the left and right hand side.

Which is to say that instead of taking `bar_format` as a kwarg and then just calling format() on the supplied string with a dict, one should either take `l_bar` and `r_bar` as kwargs, format() them and then format the bar itself with a correct width (terminal_width - `len(l_bar_fmt)` - `len(r_bar_fmt)`), or:

```python
if '{bar}' in bar_format:
    l_bar, r_bar = bar_format.split('{bar}')
    l_bar, r_bar = l_bar.format(**bar_args), r_bar.format(**bar_args)
    N_BARS = max(1, ncols - len(l_bar) - len(r_bar)) if ncols else 10
    
    # snip
    
    return l_bar + full_bar + r_bar
```

This actually allows for using `bar_format` without totally breaking the printing of that bar to the terminal, and additionally those format strings (and their de/composition) can be done inside of `__init__` somewhat obviating the need for the defaults optimizations in the first place.

karanlyons:
(This is only tangentially related, but it'd likely be worth adding cursor hiding/showing support. Under POSIX this can just be done with control sequences (`\033[?25l`/`\033[?25h`). Windows is a bit trickier.)

lrq3000:
Ok good suggestion, this is possible to do. Can I ask you just one more favour: can you provide us an example where the current custom bar formatting breaks?

lrq3000:
@karanlyons Ok I understand your code, it will be implemented. Indeed, the current code does not adapt the bar size according to what you put in `l_bar` and `r_bar`.

However, what's your idea about formatting the strings in `__init__` ? That would indeed be amazing, but since the dict content is changing at each call to the method, I can't see how we can precompute the formatting.

lrq3000:
Merged in #160.

PR #160, opened 2016-05-08
Fix bar_format breaking console display

lrq3000:
Apply the fix proposed by #149 to avoid breaking display when using `bar_format`.

This change does not add any additional computation when not using `bar_format`.

In addition, `bar_format` is now much more efficient when not using `{bar}` since it will return directly, skipping the unnecessary computations.

PS: this PR includes the changes in branch `inv-option`, sorry for that. Will clean up if `inv-option` does not get merged.

- [x] Test if it works in all cases
- [x] Coverage?


lrq3000:
This PR is done, I've got nothing to add :) I tested a few cases but I hope it will fit every cases (at least the unit tests did pass except one that had to be adapted because it relied on the non-flexible {bar} bug).

lrq3000:
Merged manually.

PR #181, opened 2016-06-07
Add callable support for bar_format argument

lrq3000:
The main goal is to allow the parent application to manage the progress bar display itself. For example, if an application with a GUI (such as [moviepy](https://github.com/Zulko/moviepy/issues/128)?) wants to use `tqdm` on both its console and GUI displays without having to make a dedicated module (that won't be reusable for any other module but theirs). This is way easier than subclassing `tqdm` and should be pretty much as fast performance-wise.

Todo:
- [ ] test on real cases (not yet tested at all, beware!)
- [ ] unit test

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -290,8 +290,9 @@ Parameters
     Exponential moving average smoothing factor for speed estimates
     (ignored in GUI mode). Ranges from 0 (average speed) to 1
     (current/instantaneous speed) [default: 0.3].
-* bar_format  : str, optional  
+* bar_format  : str or callable, optional  
     Specify a custom bar string formatting. May impact performance.
+    Can be a callable that will handle bar display.
     If unspecified, will use '{l_bar}{bar}{r_bar}', where l_bar is
     '{desc}{percentage:3.0f}%|' and r_bar is
     '| {n_fmt}/{total_fmt} [{elapsed_str}<{remaining_str}, {rate_fmt}]'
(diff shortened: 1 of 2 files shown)


casperdcl:
ah, nice... an update callback wrapped by another update callback.

lrq3000:
Ah you're right, I didn't give it enough thought, this code probably
doesn't work as it is.

I think it would be better to just provide a special value to format_meter
so that it returns the dict to status_printer, so users would just have to
overload tqdm.status_printer with their bar display code, and tqdm.__init__
to call format_meter with the special argument like bar_format=True.

The idea is to avoid forcing the user to recode the rate, remaining and
percentage computation codes that are done in format_meter. If they have a
GUI bar, they can't just overload status_printer, as
status_printer.print_status() expects a string, and a GUI usually breaks
down the different parts of the bar into different widgets.

2016-06-08 19:36 GMT+02:00 Casper da Costa-Luis <notifications@github.com>:

casperdcl:
I think it might make sense for a call to `update()` to return a dict to simplify the end interface... wasn't finding fault; thought it was a good idea.

``` python
from tqdm import tqdm

def callback_gen(t, ...):
  def inner(...):
    params = t.update(dict=true)
    # use params
  return inner

callback = callback_gen(tqdm(total=len(iter)))
for x in iter:
  callback()
```

lrq3000:
This should be reworked on to also provide the separate values when `total=None`. Also, merging this will not only help third-party devs but also our own modularization of the code, for example #223 (because then we can call parent and avoid copy/pasting the calculation code). We should then also add N_BARS as a returing argument.

lrq3000:
Closing this, this should be merged with #223 as it uses `bar_format_callback` to reduce code duplication (not really the callback but the dict return, but if dict return works then callback will also work because this is the same code but one if branching).

lrq3000:
Note that #223 contains bugfixes compared to the version in here, this is the reason I close this PR.

PR #192, opened 2016-07-03
Fix tqdm_pandas(tqdm_notebook) by delaying instanciation in tqdm_pandas

### LINKED ISSUE #182: Could the pandas progress bar be used for series objects?
If I have a very long series with an expensive function to apply, it would be good to use the pandas `.progess_apply` too. It seems that only dataframes and groupbys have this ability.

For example
```python
df = pd.DataFrame(range(1000), columns=['numbers'])
tqdm_pandas(tqdm_notebook())
df.numbers.progress_apply(lambda x: x**2)
```
Gives the following error:
``` 
AttributeError: 'Series' object has no attribute 'progress_apply'
```

lrq3000:
This fixes #182 but this breaks retrocompatibility since `tqdm_pandas` does not accept a bar instance anymore (`tqdm_pandas(tqdm(leave=True))`) but a bar class (`tqdm_pandas(tqdm, leave=True)`).

/EDIT: fixed backwards compatibility.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -497,8 +497,8 @@ for ``DataFrame.progress_apply`` and ``DataFrameGroupBy.progress_apply``:
     df = pd.DataFrame(np.random.randint(0, 100, (100000, 6)))
 
     # Create and register a new `tqdm` instance with `pandas`
-    # (can use tqdm_gui, optional kwargs, etc.)
-    tqdm_pandas(tqdm())
+    # (can use tqdm_gui, tqdm_notebook, optional kwargs, etc.)
+    tqdm_pandas(tqdm, leave=True)
 
     # Now you can use `progress_apply` instead of `apply`
     df.progress_apply(lambda x: x**2)
--- a/README.rst
+++ b/README.rst
@@ -490,15 +490,15 @@ for ``DataFrame.progress_apply`` and ``DataFrameGroupBy.progress_apply``:
 
     import pandas as pd
     import numpy as np
-    from tqdm import tqdm, tqdm_pandas
+    from tqdm import tqdm
 
     ...
 
     df = pd.DataFrame(np.random.randint(0, 100, (100000, 6)))
 
-    # Create and register a new `tqdm` instance with `pandas`
-    # (can use tqdm_gui, tqdm_notebook, optional kwargs, etc.)
-    tqdm_pandas(tqdm, leave=True)
+    # Register `pandas.progress_apply` with `tqdm`
+    # (can use `tqdm_gui`, `tqdm_notebook`, optional kwargs, etc.)
+    tqdm.pandas(desc="my bar!")
 
     # Now you can use `progress_apply` instead of `apply`
     df.progress_apply(lambda x: x**2)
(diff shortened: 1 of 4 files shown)


casperdcl:
Hmm I'd asked about doing it this way instead when creating this feature but since there didn't seem to be a clear advantage back then we stuck witht the current version... this PR is not backwards compatible, but could be fixed.

lrq3000:
Yes indeed I thought about it afterwards, we can simply detect whether the
supplied bar argument is a class or an object, and then act accordingly,
this will be backward compatible then :)
Le 5 Jul. 2016 09:07, "Casper da Costa-Luis" <notifications@github.com> a
écrit :

lrq3000:
I added a deprecation message using `tqdm.write()`, because I didn't like the `warnings` module for our purpose since it suppresses warnings by default now [as discussed here](https://code.djangoproject.com/ticket/18985).

lrq3000:
Wow that's great @casperdcl , thank's a lot!

Just out of curiosity, why did you make tqdm_pandas a classmethod? What is the advantage in this case?

casperdcl:
because the way you'd re-written it in this PR was `function(Class, ...)` which is precisely what class methods are for! The nice thing about this is all sub classes automatically inherit the class method, with the correct child class passed as the first argument when invoked.

combined with the fact that an instance is created within the inner function, this means `.pandas()` only needs to be called once in a program (more precisely, once every time you want to use a different tqdm-like *class* to handle `progress_apply`). as in:

``` python
tqdm.pandas()
... # any number of `progress_apply` calls will go to terminal
tqdm_gui.pandas()
... # now use gui instead
```

casperdcl:
re: python warning confusions, much sad such is. was vaguely aware of this; surely there's a better solution.  We could always `class TqdmDeprecationWarning(SomethingAppropriate)`?

lrq3000:
@casperdcl Great about the classmethod, talk about the low entry high ceiling of python, I still often learn new stuff XD

I thought about doing a class for the warning, but how would you raise it? If you raise it as an exception, then the old way of using tqdm_pandas will need to be specifically caught by users, so it breaks retrocompatibility. I do not know of any other way to raise a non-blocking warning in Python than just printing...

lrq3000:
But yeah it would be nice to have a standardized way of raising deprecation warnings, because I guess we will get more and more as time goes.

CrazyPython:
@lrq3000 There is! [Here's a doc link!](https://docs.python.org/3.5/library/exceptions.html#DeprecationWarning) ;)

lrq3000:
@CrazyPython Yes but the `warning` module doesn't emit any output to sys.stderr by default now [as reported here](https://code.djangoproject.com/ticket/18985) except if you modify the filters at the module level, which is considered a bad practice. I tried it with Python 2.7 just to make sure and indeed I had no output, but maybe I used it wrong or maybe the behavior was reverted in Python 3.5 (because it clearly contradicts the documentation you referenced).

lrq3000:
Ok so maybe we can add something like that in `._utils.py`:

```
class TqdmDeprecationWarning(object):
    @staticmethod
    def warn(message, file=sys.stderr):
        file.write('DeprecationWarning: %s' % message)
```

Then to use:

```
from ._utils import TqdmDeprecationWarning

TqdmDeprecationWarning.warn('parameter xxx should not be used, prefer xxx')
```

What do you think about it guys?

PR #195, opened 2016-07-17
Update test to run with python3 executable

arunpersaud:
Use sys.executable to run with the same python as the tests are run. Needed for installations with different python installed, e.g. python2 and python3 instead of plain python.

### DIFF
--- a/tqdm/tests/tests_main.py
+++ b/tqdm/tests/tests_main.py
@@ -19,7 +19,7 @@ def test_main():
     ls = subprocess.Popen(('ls'),
                           stdout=subprocess.PIPE,
                           stderr=subprocess.STDOUT)
-    res = _sh('python', '-c', 'from tqdm import main; main()',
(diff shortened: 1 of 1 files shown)


lrq3000:
Thank you for the patch, I cannot test it on my current system (Windows), but it seems all OK to me + Travis is OK (and it's running Linux).

casperdcl:
hmm I'll test this later today. I'd been working on some cli tweaks, and those tests passed on py2-linux, py2-win, py3-win, but failed on travis' py3-linux. You never know ;)

casperdcl:
ah I just looked at the actual commit. thought it involved monkeypatching or similar... this is fine ha.

lrq3000:
Yes Casper I thought the same but the patch is very clean and pythonic. I
think we can very safely merge it if you can confirm the unit tests still
work on your machine :)

2016-07-18 10:27 GMT+02:00 Casper da Costa-Luis <notifications@github.com>:

PR #196, opened 2016-07-18
Fix tqdm.write() when not yet instanciated

### LINKED ISSUE #193: tqdm work with logging
I tried "Redirecting writing" samples which works great. While I wants to use logging.info() instead of print under python 2.7, it doesn't work.
Is it possible to redirect logging just like redirect print?

Thanks in advance.

lrq3000:
Should fix #193.

Maybe should add a unit test for threads? (but is that really that useful?)

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -308,15 +308,14 @@ class tqdm(object):
 
         # Clear all bars
         inst_cleared = []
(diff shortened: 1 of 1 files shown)


lrq3000:
Thank's!

lrq3000:
And yes it's cleaner @casperdcl , I didn't even know about this functionality of getattr :+1:

PR #223, opened 2016-08-04
Custom bar symbols formatting + bar_format callback and dict return

### LINKED ISSUE #925: Progress bar with '...'
Hi, I'd love to start using this library, is there any easy way to print the typical '...' progress bar? Ie, something like this:

    import time
    import threading
    import sys

    class ProgressBar():

        def __init__(self):
            self.start = time.time()
            self.count = 0
            self.completed = False
            self.width = 4

        def stop(self):
            self.completed = True

        def print(self, msg):
            while not self.completed:
                elapsed = time.time() - self.start
                if elapsed > 0.3:
                    self.start = time.time()
                    self.count += 1
                    suffix = '.' * (self.count % self.width)
                    suffix = suffix.ljust(self.width - self.count % self.width, ' ')
                    print("{}{}".format(msg, suffix), end='\r', flush=True)

    if __name__ == '__main__':
        pb = ProgressBar()
        t = threading.Thread(target=pb.print, args=("Doing stuff",))
        t.start()
        time.sleep(5)
        pb.stop()
        t.join()

lrq3000:
Implements also #181 (add callback support for bar_format) + custom bar symbols formatting as a submodule.

Not the most urgent PR, but we missed some more customization (pretty sure someone will ask someday!). So here it is: the ability to set custom bar symbols, just like our fellow [progressbar2](https://github.com/WoLpH/python-progressbar):

```
import time
from tqdm import tcrange, tqdm_custom

bar_format = {'template': '{l_bar}{bar}{r_bar}',
                     'symbols': {'ascii': list('ABCDEO'),
                                      'loop': False},
                     }

bar_format_loop = {
                     'symbols': {'ascii': list('/-\|'),
                                      'loop': True},
                     }

bar_format_ind_loop = {
                    'symbols_indeterminate': {'ascii': list('/-\|'),
                                      'loop': True},
                      }

bar_format_ind = {
                    'symbols_indeterminate': {'ascii': ['----{,_,">', '----{,_,*>'],
                                      'loop': False},
                      }

for i in tcrange(100, ascii=True, bar_format=bar_format):
    time.sleep(0.1)

for i in tcrange(100, ascii=True, bar_format=bar_format_loop):
    time.sleep(0.1)

with tqdm_custom(ascii=True, bar_format=bar_format_ind_loop) as tt1:
    for i in xrange(100):
        time.sleep(0.1)
        tt1.update()

with tqdm_custom(ascii=True, bar_format=bar_format_ind) as tt2:
    for i in xrange(100):
        time.sleep(0.1)
        tt2.update()
```

Respective outputs:

```
  9%|OOOE                                      | 9/100 [00:00<00:09,  9.95it/s]

 12%|/| 12/100 [00:01<00:08, 10.00it/s]

100it|/| 100/? [00:10,  9.91it/s]

100it|                          ----{,_,">           | 100/? [00:10, 10.00it/s]
```

For multiline (work in progress, needs bugfixes):

```
import time
from tqdm import tcmrange, tqdm_custommulti
from copy import deepcopy

bar_format_ind_multi = {
                    'symbols_indeterminate': {'ascii': [
"""
     _
\. _(9>
 \==_)
  -'=
""",
"""
     _
\. _(6>
 \==_)
  -'=
"""
],
                                      'loop': False,
                                      'multiline': True},
                      }


bar_format_ind_loop_multi = {
                    'symbols_indeterminate': {'ascii': [
"""
     _
\. _(9>
 \==_)
  -'=
""",
"""
     _
\. _(6>
 \==_)
  -'=
"""
],
                                      'loop': True,
                                      'multiline': True},
                      }

bar_format_loop_multi = {
                    'symbols': {'ascii': [
"""
     _
\. _(9>
 \==_)
  -'=
""",
"""
     _
\. _(6>
 \==_)
  -'=
"""
],
                                      'loop': True,
                                      'multiline': True},
                      }

bar_format_multi = {
                    'symbols': {'ascii': [
"""
. . . 
 . .
. . .
""",
"""
* * * 
 * *
* * *
""",
"""
= = = 
 = =
= = =
""",
"""
0 0 0 
 0 0
0 0 0
""",
"""
O O O 
 O O
O O O
""",
                                    ],
                                      'loop': False,
                                      'multiline': True},
                      }

bar_format_multi_anim = {
                    'symbols': {'ascii': [
["""
. . . 
 . .
. . .
""",
"""
 . .
. . . 
 . .
"""],
["""
* * * 
 * *
* * *
""",
"""
 * *
* * * 
 * *
"""],
["""
= = = 
 = =
= = =
""",
"""
 = =
= = = 
 = =
"""],
["""
0 0 0 
 0 0
0 0 0
""",
"""
 0 0
0 0 0 
 0 0
"""],
["""
O O O 
 O O
O O O
""",
"""
 O O
O O O 
 O O
"""],
                                    ],
                                      'loop': False,
                                      'multiline': True,
                                      'random': False},
                      }

bar_format_multi_anim_rand = deepcopy(bar_format_multi_anim)
bar_format_multi_anim_rand['symbols']['random'] = True

sleep_step = 0.05
anim_interval = 0.1

for i in tcmrange(100, mininterval=anim_interval, ascii=True, bar_format=bar_format_loop_multi):
    time.sleep(sleep_step)

with tqdm_custommulti(mininterval=anim_interval, ascii=True, bar_format=bar_format_ind_loop_multi) as ttm1:
    for i in xrange(100):
        time.sleep(sleep_step)
        ttm1.update()

with tqdm_custommulti(mininterval=anim_interval, ascii=True, bar_format=bar_format_ind_multi) as ttm1:
    for i in xrange(300):
        time.sleep(sleep_step)
        ttm1.update()

for i in tcmrange(100, mininterval=anim_interval, ascii=True, bar_format=bar_format_multi):
    time.sleep(sleep_step)

for i in tcmrange(100, mininterval=anim_interval, ascii=True, bar_format=bar_format_multi_anim):
    time.sleep(sleep_step)

for i in tcmrange(100, mininterval=anim_interval, ascii=True, bar_format=bar_format_multi_anim_rand):
    time.sleep(sleep_step)
```

User can specify both ascii and unicode versions of the bar, if one isn't provided, then it will fallback to the default tqdm bar.

TODO:
- [x] Add indeterminate progress customization
- [x] Optimize reversing symbol by precomputing in `__init__` and then loading it at the same time as `c_symbol` (eg, `c_symbol_reverse = self.bar_format[key].get('ascii_reverse', None)` ? Would allow manual user reversing BTW).
- [x] Rebase on #181 
- [x] Remove copy/pasted code thank's to #181 
- [x] Multiline symbols support (fish/nyanbar-like, see below)
  - [x] Multiline indeterminate bar
  - [x] Multiline loop (both indeterminate and determinate)
  - [x] Multiline animated progress bar
  - [ ] Fix leave=True (there are more than the required line returns sometimes, I guess it's Windows terminal's fault, it seems it doubles the number of line returns when reaching the terminal's end)
  - [ ] Fix multiline progress bar overflowing ncols when 100%
  - [ ] Fix multiline progress bar frac not reaching filler symbol when 100%
  - [x] Fix multiple parallel/nested multiline bars using `self.last_print_height` (works with leave=False, need to fix leave=True bug)
- [ ] Fix Flake8
- [ ] Add unit tests
- [ ] Update docstrings and readme

About multiline symbols:
We should try to implement multiline animated ascii-art bars like [nyanbar](https://github.com/apg/nyanbar) or [fish](https://github.com/lericson/fish). This shouldn't be too hard to implement, let me explain:

Basically, we have currently 4 different bar displays:
- **progress bar** : standard progress bar with numbers 1 to 9 and then `#` when enough progress, and it gets filled up until it reaches 100% completion.
- **progress loop** : animated symbol that loops in-place, takes minimal space on screen.
- **indeterminate bar** : infinite progress bar with a symbol that goes from left to right, then right to left (because we don't know when it ends).
- **indeterminate loop** : like progress loop.

Then for all these cases, you can add `multiline` support. We can then assimilate `fish` and `nyanbar` to the following cases:
- fish single line = indeterminate bar
- fish multi line = indeterminate bar **multiline**
- nyanbar = progress bar **multiline**

So if we can support multiline, we can essentially support what these two libraries do (somewhat, because for example `nyanbar` would only be simulated since the real one generates random colored blocks, our version would have fixed sprites but if you make enough of them, then you can give the illusion it's random).

Only issue is how to handle multiple concurrent tqdm bars, but this can be managed I think (we can memorize the number of lines of the last printed symbol and get back to original position, just like we currently do with automated nesting, then when we `move_to()` we would compute the correct position to print using `sum([t.s_height for t in tqdm._instances if t.pos < self.pos])`).

---

**The rest of the text here is for the old version of this PR.**

For reference, the old way with bar string templating was like this (but it was scraped to give more flexibility, but maybe it will be reintroduced in the future if people like it):

```
import time
from tqdm import tcrange

for i in tcrange(100, ascii=True, bar_format='{l_bar}{bar}{bar_symbols_ascii},A,B,C,D,E,O{/bar_symbols_ascii}{r_bar}'):
    time.sleep(0.1)

for i in tcrange(100, ascii=True, bar_format='{l_bar}{bar}{bar_symbols_loop_ascii},/,-,\,|{/bar_symbols_loop_ascii}{r_bar}'):
    time.sleep(0.1)
```

There are four different new tags for `bar_format`:
- `{bar_symbols}` and `{bar_symbols_ascii}` set custom progress bar symbols (unicode or ascii respectively)
- `{bar_symbols_loop}` and `{bar_symbols_loop_ascii}` set animated looping symbols (unicode or ascii respectively)

As you can see, I added both unicode and ascii tags, so that the user can provide either one or both. If the user didn't provide custom symbols for the target environment (eg, user provided `{bar_symbols}` but the environment is ascii), then the bar will fallback to the default bar. Why? This is to ensure robustness, so the user has to explicitly provide both a unicode and an ascii bars to work in all envs.

Formatting follow a rule similar to egrep: the first character is the separator, then the rest is compiled in a list of symbols. Eg:
- `{bar_symbols}|1|2|3|#{/bar_symbols}` --> `|` is the separator
- `{bar_symbols_loop},|,/,-,\{/bar_symbols_loop}` --> `,` is the separator

Note that symbols are not necessarily only one character long, they can be longer (can be interesting for animated loops). /EDIT: tested and it works.

### DIFF
--- a/tqdm/_tqdm_custom.py
+++ b/tqdm/_tqdm_custom.py
@@ -22,6 +22,15 @@ __author__ = {"github.com/": ["lrq3000"]}
 __all__ = ['tqdm_custom', 'tcrange']
 
 
+def mirror_chars(s):
+    """Mirror characters using translation"""
+    import string
+    ins = '()<>[]\/{}bd'
+    outs = ')(><][/\}{db'
+    trans = string.maketrans(ins,outs)
+    return s.translate(trans)
+
+
 class tqdm_custom(tqdm):
     """
     tqdm with nice customizable bar symbols!
@@ -229,8 +238,54 @@ class tqdm_custom(tqdm):
 
         # no total: no progressbar, ETA, just progress stats
         else:
-            return (prefix if prefix else '') + '{0}{1} [{2}, {3}]'.format(
-                n_fmt, unit, elapsed_str, rate_fmt)
+            if bar_format and isinstance(bar_format, dict):
+                # left and right sides
+                l_bar = '{0}{1}|'.format(n_fmt, unit)
+                r_bar = '| {0}/? [{1}, {2}]'.format(n_fmt, elapsed_str, rate_fmt)
+
+                # space available for bar's display
+                N_BARS = max(1, ncols - len(l_bar) - len(r_bar)) if ncols \
+                    else 10
+
+                # get ascii or unicode template
+                if ascii:
+                    c_symb = bar_format['symbols_indeterminate'].get('ascii', ["====="])
+                else:
+                    c_symb = bar_format['symbols_indeterminate'].get('unicode', ["====="])
+                # looping symbols: just update the symbol animation at each iteration
+                if bar_format['symbols_indeterminate'].get('loop', False):
+                    # increment one step in the animation at each step
+                    bar = c_symb[divmod(n, len(c_symb))[1]]
+
+                    bar_length = N_BARS  # avoid the filling
+                # indeterminate progress bar (cycle from left to right then right to left)
+                else:
+                    # Get current bar animation based on current iteration
+                    bar = c_symb[divmod(n, len(c_symb))[1]]
+                    # Get left filling space and animation step (right pass or left?)
+                    anim_step, fill_left = divmod(n, (N_BARS - len(bar)))
+                    # If anim_step is odd, then we do left pass (2nd pass)
+                    if divmod(anim_step, 2)[1] == 1:
+                        # Inverse the left filling space (now it's the right space)
+                        fill_left = N_BARS - len(bar) - fill_left
+                        # Reverse the bar string
+                        bar = mirror_chars(bar[::-1])
+
+                    # Generate bar with left filling space
+                    bar = ' ' * fill_left + bar
+                    bar_length = len(bar)
+
+                # Right space padding
+                full_bar = bar + \
+                    ' ' * max(N_BARS - bar_length, 0)
+
+                # Piece together the bar parts
+                return l_bar + full_bar + r_bar
+
+            # Standard fast no ETA progress status
+            else:
+                return (prefix if prefix else '') + '{0}{1} [{2}, {3}]'.format(
+                        n_fmt, unit, elapsed_str, rate_fmt)
 
     def __init__(self, *args, **kwargs):
         """
--- a/tqdm/_tqdm_custom.py
+++ b/tqdm/_tqdm_custom.py
@@ -35,8 +35,7 @@ class tqdm_custom(tqdm):
     """
     tqdm with nice customizable bar symbols!
     """
-    @staticmethod
-    def format_meter(n, total, elapsed, ncols=None, prefix='',
+    def format_meter(self, n, total, elapsed, ncols=None, prefix='',
                      ascii=False, unit='it', unit_scale=False, rate=None,
                      bar_format=None):
         """
@@ -187,8 +186,10 @@ class tqdm_custom(tqdm):
                     c_symb = bar_format['symbols'].get('unicode', map(_unich, range(0x258F, 0x2587, -1)))
                 # looping symbols: just update the symbol animation at each iteration
                 if bar_format['symbols'].get('loop', False):
-                    # increment one step in the animation at each step
-                    bar = c_symb[divmod(n, len(c_symb))[1]]
+                    # increment one step in the animation for each display
+                    self.n_anim += 1
+                    # get the symbol for current animation step
+                    bar = c_symb[divmod(self.n_anim, len(c_symb))[1]]
                     frac_bar = ''
 
                     bar_length = N_BARS  # avoid the filling
@@ -254,16 +255,20 @@ class tqdm_custom(tqdm):
                     c_symb = bar_format['symbols_indeterminate'].get('unicode', ["====="])
                 # looping symbols: just update the symbol animation at each iteration
                 if bar_format['symbols_indeterminate'].get('loop', False):
-                    # increment one step in the animation at each step
-                    bar = c_symb[divmod(n, len(c_symb))[1]]
+                    # increment one step in the animation for each display
+                    self.n_anim += 1
+                    # Get current bar animation based on current iteration
+                    bar = c_symb[divmod(self.n_anim, len(c_symb))[1]]
 
                     bar_length = N_BARS  # avoid the filling
                 # indeterminate progress bar (cycle from left to right then right to left)
                 else:
+                    # increment one step in the animation for each display
+                    self.n_anim += 1
                     # Get current bar animation based on current iteration
-                    bar = c_symb[divmod(n, len(c_symb))[1]]
+                    bar = c_symb[divmod(self.n_anim, len(c_symb))[1]]
                     # Get left filling space and animation step (right pass or left?)
-                    anim_step, fill_left = divmod(n, (N_BARS - len(bar)))
+                    anim_step, fill_left = divmod(self.n_anim, (N_BARS - len(bar)))
                     # If anim_step is odd, then we do left pass (2nd pass)
                     if divmod(anim_step, 2)[1] == 1:
                         # Inverse the left filling space (now it's the right space)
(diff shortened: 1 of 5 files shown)


lrq3000:
PS: I forgot to say that I tried to minimize the impact of this PR on standard core `tqdm` (without `bar_format` supplied): there are two variables instanciations at the top and then one additional if statement. That's all. The rest is entirely done in `bar_format` if branch and inside the new `c_symbols` if branch.

casperdcl:
I was just working on this! going to merge into this PR... thanks, Stephen.

lrq3000:
Ah lol @casperdcl, maybe some kind of telepathy was at work here? XD

lrq3000:
I didn't plan to work on this at all, it just popped into my mind a few days ago, so that's indeed quite a coincidence we were working on the same feature at the same time :)

lrq3000:
Just for reference and for fun:
* https://github.com/iarna/gauge/issues/5#issuecomment-146386489
* http://changaco.oy.lc/unicode-progress-bars/ (from https://github.com/tqdm/tqdm/issues/50#issuecomment-155552898)
* https://github.com/tqdm/tqdm/issues/38#issuecomment-147491456

CrazyPython:
@lrq3000 to minimize performance strain, let's make a branch that internally changes the code, or even make a new class. This small feature shouldn't slow everybody down.

lrq3000:
For standard tqdm, it's an increase of 3 opcodes per printing (not per loop), this doesn't impact much the performance.

However it can probably be implemented in a separate submodule. It would incur repetition of some code (calculations of the bar length and current frac) and the change of the tags from `{bar_format}` to `[bar_format]` which is a little bit weird because it will be different than other tags, but necessary else the formatting will choke.

lrq3000:
Fixed the custom_symbols tag string extraction, now it's preprocessed once in `tqdm.__init__()`. The performance impact is 1 additional if statement in `format_meter()` with `bar_format=None`, and it's roughly the same as before custom_symbols for users specifying a template in `bar_format` (the calculations about the bar size has a few more operations but it's constant time so it's ok).

lrq3000:
Moved all the code into its own module in `tqdm_custom`, so we can freely experiment and implement things like #228 without fearing that it will slow things down. If later we find a way to implement these features without overhead, we can move them back to the core.

Please do NOT squash all the commits, because the original commit can be useful to reimplement back to the core.

lrq3000:
Added two new features:

* customization of indeterminate progress status (two ways: looping symbol or looping bar similarly to fish).
* looping speed is not dependent on `n` anymore but on `mininterval`. So this means that `mininterval` sets the maximum speed, not the minimum speed (if some iterations take longer than usual, if the bar is not refreshed then the animation won't move, that's normal because `tqdm` refresh display only when necessary, there's no threading, and it allows for great efficiency).

tqdm/_tqdm_custom.py
```diff
@@ -0,0 +1,263 @@
...
+                    # Get current bar animation based on current iteration
+                    symbol_idx = divmod(self.n_anim, len(c_symb))[1]
+                    bar = c_symb[symbol_idx]
+                    # Get left filling space and animation step (right pass or left?)
+                    anim_step, fill_left = divmod(self.n_anim, (N_BARS - len(bar)))
+                    # If anim_step is odd, then we do left pass (2nd pass)
+                    if divmod(anim_step, 2)[1] == 1:
+                        # Inverse the left filling space (now it's the right space)
+                        fill_left = N_BARS - len(bar) - fill_left
+                        # Get the reversed symbol
```
CrazyPython: @lrq3000 🚨 🚨inconsistent comment capitalization

there should be a linter for this.

CrazyPython:
@lrq3000 I've seen you use "thank's" on two ocassions now. This is just a friendly grammar suggestion, so here goes:

 1. It is "thanks".
 2. `thank's` is the possessive form of `thank`, which raises this exception:
```
TypeError: Verb object does not support possessive operations.
```

then I went into a coma for four hours until a script restarted me. Until I read this again. Here I am telling you not to make me go into a coma.

lrq3000:
Ok I'm fed up with this PR, so I'll leave it aside for now.

It's already working pretty well, and it has gone further than what I intended at first. There are only a few remaining bugs in multiline mode, such as `leave=True` doubling line returns. If someone has any idea to fix that, that'd be great (or if someone can just test on Linux and tell me if you have the same issue).

Please feedback if you can find any use to this module. Nice ASCII animations to showcase are welcome.

lrq3000:
The bar_format callback should be separated back into its own PR and merged with milestone >5.

rth:
@lrq3000 Thanks for this PR (and tqdm in general )!

> The bar_format callback should be separated back into its own PR and merged with milestone >5.

Is anyone planning to work on this part in the near future ? If not, I could give it a try..

casperdcl:
Might merge in #227 (which is simple) and put this PR into a submodule as suggested there as it's quite large

PR #239, opened 2016-08-16
Get rid of exception when closing tqdm_notebook without a total

aplavin:


tqdm/_tqdm_notebook.py
```diff
@@ -205,7 +205,7 @@ def close(self, *args, **kwargs):
         super(tqdm_notebook, self).close(*args, **kwargs)
         # Try to detect if there was an error or KeyboardInterrupt
         # in manual mode: if n < total, things probably got wrong
-        if self.n < self.total:
+        if self.total is not None and self.n < self.total:
```
lrq3000: Thank you @aplavin !

Can you please just change the line to:

```
if self.total and self.n < self.total:
```

(except if there's a specific reason you want to check `is not None`)

lrq3000:
Perfect, thank's a lot ! :+1:

PR #244, opened 2016-08-18
Fix total computation for pandas apply

aplavin:
Note, this is still not an always-correct solution! May be we need to have separate logic for apply/map/applymap?..

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -402,8 +402,13 @@ class tqdm(object):
                 # Precompute total iterations
                 total = getattr(df, 'ngroups', None)
                 if total is None:  # not grouped
-                    axis = kwargs.get('axis', 0)
-                    total = df.shape[axis]
+                    if df_function == 'applymap':
+                        total = df.size
+                    else:
+                        axis = kwargs.get('axis', 0)
+                        total = df.shape[axis]
+                        if isinstance(df, DataFrame):
+                            total += 1  # pandas calls update once too many
                 else:
                     total += 1  # pandas calls update once too many
 
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -402,14 +402,8 @@ class tqdm(object):
                 # Precompute total iterations
                 total = getattr(df, 'ngroups', None)
                 if total is None:  # not grouped
-                    if isinstance(df, Series):
-                        total = len(df)
-                    elif kwargs.get('axis', 0) == 0:
-                        # over columns
-                        total = df.size // len(df)
-                    else:
-                        # over rows
-                        total = len(df)
+                    axis = kwargs.get('axis', 0)
+                    total = df.shape[axis]
                 else:
                     total += 1  # pandas calls update once too many
 
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -388,7 +388,7 @@ class tqdm(object):
         deprecated_t = [tkwargs.pop('deprecated_t', None)]
 
         def inner_generator(df_function='apply'):
-            def inner(df, func, *args, **kwargs):
+            def inner(df, func, **kwargs):
                 """
                 Parameters
                 ----------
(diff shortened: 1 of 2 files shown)


tqdm/_tqdm.py
```diff
@@ -402,8 +402,14 @@ def inner(df, func, *args, **kwargs):
                 # Precompute total iterations
                 total = getattr(df, 'ngroups', None)
                 if total is None:  # not grouped
-                    total = len(df) if isinstance(df, Series) \
-                        else df.size // len(df)
+                    if isinstance(df, Series):
+                        total = len(df)
+                    elif kwargs.get('axis', 0) == 0:
```
aplavin: `axis` may also be in `args` instead of `kwargs`... how to support it cleanly?

tqdm/_tqdm.py
```diff
@@ -402,8 +402,14 @@ def inner(df, func, *args, **kwargs):
...
+                        # over columns
+                        total = df.size // len(df)
+                    else:
+                        # over rows
```
aplavin: assumes that `axis == 1`, according to pandas documentation it must be the case for `DataFrame` (but maybe someday `tqdm` will support `Panel`s)

tqdm/_tqdm.py
lrq3000: Simply:
```
elif kwargs.get('axis', 0) == 0 or args[0] == 0:
```

Because the other arguments are not integers. But yeah it's not foolproof, but I cannot see any better way to handle that.

tqdm/_tqdm.py
lrq3000: Actually we can probably already support panels, it's just that @casperdcl added pandas features per user requests. Panels are quite similar to DataFrame so it should work as-is I guess.

tqdm/_tqdm.py
aplavin: Yes, not foolproff, but I agree there is no other simple way.

tqdm/_tqdm.py
aplavin: I mean there must be another axis for panels, never used them myself.

aplavin:
And this certainly still doesn't work for `applymap` for example: it needs `total = df.size`.

tqdm/_tqdm.py
aplavin: And after thinking a bit more, I would say we should either make separate functions for separate `pandas` ones, or disallow `*args` completely. For example, `Series.apply(func, convert_dtype=True, args=(), **kwds)` has the first argument as `bool` which can be passed as `0`, and this can lead to very subtle bugs.

lrq3000: I agree that we could disallow *args completely. We should just output a clear message that positional arguments are not supported with tqdm_pandas. This would be the best I think.

tqdm/_tqdm.py
lrq3000: Neither did I XD Yes there must more axes but it's just the same argument I think, so it's the same management (just use `panel.shape[x]` where x is the axis if I'm not mistaken).

aplavin: Well, we can just use `shape` in all cases :) See later commit.

aplavin:
Well, there are other subtleties: for example, one can pass named axis (e.g. `axis = 'index'`), or for panels even two axes as a tuple. This patch doesn't support it.

lrq3000:
Ah yes in this case it will be bothersome to handle all these special cases...

lrq3000:
@casperdcl What do you think of this PR? Can we merge it in?

CrazyPython:
`args not supported intentionally` -> `args intentionally not supported`

lemme send PR

lrq3000:
Thank's @CrazyPython :)

aplavin:
Any updates on merging this?..

CrazyPython:
@aplavin merge typo PR

lrq3000:
Casper seems to be busy and I'm moving to another country so I don't yet
have a stable Internet connection, but by the end of october it will surely
be merged, along with lots of other PRs :)

CrazyPython:
- [ ] Add tests
- [ ] Get reviewed

casperdcl:
in #299.

Issue #206, opened 2016-07-21
Regression prediction mode


CrazyPython:
I had a situation where every iteration, the time of the iteration slowed down. I propose a keyword argument to regress the time of each iteration to have better prediction time. It could simply do a exponential and linear regression and see which one fits better.

Like: `tqdm(..., regressmode=True)`

If this is to be implemented, there should be a warning about the possible slowdown.

lrq3000:
This case is difficult and rare. The problem in implementing regressive methods is that they need to store all the past states (or at least a window), and `tqdm` philosophy of being fast and efficient implies that we use memoryless methods (we only know one saved value, the average, and the current value).

We talked about implementing an optional predictive/regressive model, but it is a great endeavour, and quite innovative (I didn't see any paper or software implementing such a model for progress bars or time prediction).

Initially, we intended to try to use machine learning, but it may be a bit too much.

Now I'm wondering if maybe we could use a memoryless regressive method, but I never heard of such a class of algorithm, a google scholar search should clarify that.

I'm now thinking such a memoryless estimator is possible because, when we think about it, what `tqdm` currently does is to estimate the average/min/max bounds using EMA.

Instead, we could maybe try to compute the vector of time progress, so we could know not only the value (how long it will take for next iteration), but also the direction (how much it will impact next next iteration time: constant, increasing or decreasing?).

If someone has heard of any algorithm that could fit here, I'd be glad to hear about it!

CrazyPython:
@lrq3000 `tqdm.regress` as a separate one, so it doesn't slow down main operations? My case only had 24 iterations but it would take above 30 minutes (the first one is <1 minute). You can add a warning that such function is only for small iteration numbers. 

Just use the kind of simple regress method - the kind you find on a TI-84 calculator. and TI-84 has like 2 kHZ speed.

casperdcl:
now we're talking. decent math rather than all this recent python implementation detail. Second order ema? Sounds v. promising, gonna give it a go this weekend :)

casperdcl:
also, ti-84(+)? hello fellow ib learner!

lrq3000:
Lol yeah I remember you are quite into data science @casperdcl :) I have absolutely no idea how a second order ema could be implemented, but yeah that would pretty much be the idea if you can do it :)

Otherwise, I have been thinking a bit about it. If we plot the cumulative time (current total time) as y and the number of iterations as x, we get something like this (warning: mspaint ahead):

[...]

PR #248, opened 2016-08-26
Polynomial regression to predict time (tqdm_regress, tsrange)

lrq3000:
Add tqdm module to predict time using machine learning polynomial regression.

Implementation of what was asked in #206.

For memory, here is the model used in this implementation to predict time using machine learning polynomial regression:

image

Canonical example:

```
import time
from tqdm import tsrange

plot = True
tstep = 0.1

tt = tsrange(100, leave=True, miniters=1, mininterval=0, algo='numpy', order=3, plot=plot)
#tt = tsrange(100, leave=True, miniters=1, mininterval=0, algo='descent', repeat=30, plot=plot)
#tt = tsrange(100, leave=True, miniters=1, mininterval=0, algo='sklearn', order=3, repeat=30, plot=plot)

for i in tt:
    time.sleep(tstep)
    #tstep = tstep * 1.1
    tstep = tstep + 0.01

    # DEBUG
    predicted_endtime = float(tt.model.predict([tt.total]))
    tt.write('it: %i, Predicted endtime: %g vs current: %g' % (tt.n, predicted_endtime, tt.model.predict([tt.n])))
```

Uncomment the other `tt = ...` lines if you want to try the different algorithms provided.

There are three different algorithms provided:
- **Numpy**: this is the analytical least square regression, so it always finds the best fit, but it can be slower if there are lots of samples to regress. Plus it requires numpy. This is the default, as it gives the best results currently. Plus I limited the number of samples used to fit, so the dimensionality shouldn't be an issue.
- **Descent**: pure python implementation of minibatch gradient descent for polynomial regression. It was initially set to be the default, but the analytical approach works way better in our case.
- **Sklearn**: use scikit-learn to compute a solution (iteratively, not analytically). It's similar to "descent" but it requires scikit-learn and should converge a bit faster and be more resilient against noisy samples (because it uses PassiveAggressive regressor by default, but it can be switched to stochastic gradient descent).

Ok so basically how it works:
- At each iteration, a new sample `(n=current iteration, elapsed)` is generated.
- This sample gets fuzzily added to a batch of memorized samples in the model. Why fuzzily? This isn't the best term to describe it, but basically new points are not necessarily memorized, there's a probability to store them or just to skip them. The goal was to ensure we remember old points as well as new points, but not just only new points. This should generate a better model. Why do we memorize a batch of samples? I tried with stochastic gradient descent, but with only one point the fitting gets it all wrong, we need several different samples to fit correctly, hence why we need to memorize some points, but not all else we will get out of memory if the iterable total > 10^7.
- The model learns in batch on all memorized samples at each iteration. So it adapts itself fully at each iteration to the new shape of the curve.
- Ending time is predicted using the model.
- We compute the expected average rate given the predicted ending time.
- We feed everything to format_meter.

Note on the implementation: all this is done in `format_meter()` for two reasons:
1. Easier and cleaner to implement, else we would need to copy the whole code of `__iter__()` and `update()`.
2. Lighter on CPU, because we add samples/fit only when printing, which is regulated by miniters and such. So the CPU impact of the regression is greatly lowered, at the expense of missing a few points that could help converge faster.

Please try it out and tell me if it fits your need and if it works well (don't hesitate to play with the parameters, they are explained in the docstring of `tqdm_regress`). I will finish the TODO after received feedback.

TODO:
- [ ] Flake8
- [ ] Unit tests (use virtual timer and canonical cases: addition, multiplication, subtraction, etc.)
- [ ] Readme

### DIFF
--- a/tqdm/__init__.py
+++ b/tqdm/__init__.py
@@ -2,13 +2,15 @@ from ._tqdm import tqdm
 from ._tqdm import trange
 from ._tqdm_gui import tqdm_gui
 from ._tqdm_gui import tgrange
+from ._tqdm_regress import tqdm_regress
+from ._tqdm_regress import tsrange
 from ._tqdm_pandas import tqdm_pandas
 from ._main import main
 from ._version import __version__  # NOQA
 from ._tqdm import TqdmTypeError, TqdmKeyError, TqdmDeprecationWarning
 
 __all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange', 'tqdm_pandas',
-           'tqdm_notebook', 'tnrange', 'main',
+           'tqdm_notebook', 'tnrange', 'tqdm_regress', 'tsrange', 'main',
            'TqdmTypeError', 'TqdmKeyError', 'TqdmDeprecationWarning',
            '__version__']
 
--- a/tqdm/_tqdm_regress.py
+++ b/tqdm/_tqdm_regress.py
@@ -453,7 +453,7 @@ class RegressionSklearn(RegressionNormalEq):
 
 class tqdm_regress(tqdm):
     """
-    Experimental IPython/Jupyter Notebook widget using tqdm!
+    Experimental tqdm with machine learning polynomial regression to predict time!
     """
 
     # not a static method anymore
--- a/tqdm/_tqdm_regress.py
+++ b/tqdm/_tqdm_regress.py
@@ -106,7 +106,7 @@ class TPlot(object):
 
 class RegressionPoly(object):
     """Multivariable linear/polynomial/ridge regression in pure Python"""
-    def __init__(self, total, order=2.5, batch_size=10, proba_add_batch=0.5):
+    def __init__(self, total, order=2.5, batch_size=10, proba_add_batch=0.5, norand=False):
         self.total = total
         self.order = order
         self.batch_size = batch_size
@@ -117,11 +117,15 @@ class RegressionPoly(object):
         self.last_cost = None
         self.cost_history = []
         self.last_it = 0
+        self.norand = norand
 
         # feature scale total
         self.total_poly = self.polynomize(self.total, self.order)
         # init parameters
-        self.params = [random.random() for _ in range(len(self.total_poly)+1)]  # +1 for bias unit
+        if norand:
+            self.params = [0] * (len(self.total_poly)+1)  # +1 for bias unit
+        else:
+            self.params = [random.random() for _ in range(len(self.total_poly)+1)]  # +1 for bias unit
 
     def add_sample(self, x, y, iteration=0, random_insert=True):
         self.last_it = iteration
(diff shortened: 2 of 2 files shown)


lrq3000:
PS: if you are wondering whether we could implement analytical regression with numpy, the answer is: not by ourselves, way too much complicated because we would have to reimplement a lot of linear algebra on matrices. There is however a pure python library called [PyLA](https://github.com/dmishin/pyla) that did all that and even features a least square linear regression function, so we could use that, but we would still rely on a third-party library (although pure python) and it would be a lot slower than numpy.

lrq3000:
Ah maybe we could implement an additional algorithm (Moving average regression model) as pointed here: https://github.com/tqdm/tqdm/issues/48#issuecomment-155596815

CrazyPython:
@lrq3000 By default the "regression mode" is linear. This one implements polynomial - perhaps exponential should also be implemented?

lrq3000:
@CrazyPython Exponential is implemented, just use `order=1.5` and it will compute linear, exp, sqrt and log (yes because log can always be useful in machine learning ;) ). You can also set linear mode by setting `order=1`, but from my tests it's not very useful to predict time (just a little better sometimes than our current implementation but still way off in exponentially increasing cases).

CrazyPython:
@lrq3000 oh linear = regular tqdm. put it in README?

lrq3000:
@CrazyPython Not exactly, normal `tqdm` is exponential moving average, so it cannot predict at all when iterations are taking longer and longer because it assumes constance. Linear regression is with `tqdm_regress(order=1)`, it can predict a bit more in this scenario but generally the result is not a lot better than core `tqdm`. The prediction becomes more interesting with `order=2` and the best i think with `order=3`. I tried exp, sqrt and log but they usually do not help much and generally tend to overfit.

CrazyPython:
Tag help-wanted for unit tests.

lrq3000:
@CrazyPython No need, I can make them, I can recycle my testing code into unit tests + already done unit tests for machine learning libraries, so it won't be too hard, but it's time consuming, so I'll do that only if this module has an audience.

CrazyPython:
@lrq3000 aye, is linting really so hard that you have to put it on the TODO? My editor (pcyharm) has it built-in w/ a simple keyboard shortcut.

lrq3000:
I don't use an IDE but a simple text editor notepad++, and my default font
is not monospace (didn't have the time to fix that). My main pb is with
spaces usually.

CrazyPython:
@lrq3000 Oh god you have no idea the joys of PyCharm. I used to use IDLE (notepad + indentation + shortcuts for running and embedded interactive shell), and looking back, it was pretty horrible.

PyCharm is a really "smart" IDE. If you use a `__future__` statement, it'll ask you if you want to enable code compatibility inspection. There are two-click operations for converting to list comprehension and switching between argument types. They are easy to access to. PEP 8 violations are highlighted inline, and so are type violations. Ctrl+Shift+L lints your code, applying necessary changes.

PyCharm Community Edition is free and open source. Inspections can be disabled/enabled at will. Easy to start using.

casperdcl:
fixes #206, should go into a `contrib` submodule as per https://github.com/tqdm/tqdm/pull/198#issuecomment-394184136

PR #251, opened 2016-08-28
Monitoring thread to prevent tqdm taking too much time for display

lrq3000:
Added a monitoring thread that will check all tqdm instances regularly and reduce miniters if necessary. This should fix issues like #249 and #54 for good.

About performance impact:
- This solution adds a bit of additional overhead at the first tqdm instanciation because it creates a thread, but the bright side is that then the thread sleeps most of the time, and wakes up only to check every instances' miniters value and last printing time. By default, it wakes up every 10 seconds. So the monitoring thread 
- But the bad side is that this solution needs to access and modify `self.miniters`. No problem here, the problem is with `tqdm.__iter__()` since it doesn't do any access to self variables inside loops. So I had to modify the loop to access `self.miniters`. My reasoning is that accessing `self.miniters` will use less CPU than calling `time.time()`, but I didn't try it out. If someone wants to profile to see the impact of `self.miniters`, that would be great!

/EDIT: profiled, here are the results:

**Normal tqdm (pre-PR)**:

```
for i in tqdm.trange(int(1e8)):
    pass
100%|#######################| 100000000/100000000 [00:20<00:00, 5000000.00it/s]
(20.002348575184286, 'seconds')
```

**Normal tqdm no miniters**:

```
for i in tqdm.trange(int(1e8), miniters=1, mininterval=0.1):
    pass
100%|#######################| 100000000/100000000 [00:40<00:00, 2499937.51it/s]
(40.00375434790657, 'seconds')
```

**Monitoring thread PR**:

```
for i in tqdm.trange(int(1e8)):
    pass
100%|#######################| 100000000/100000000 [00:22<00:00, 4423995.78it/s]
(22.605596624972463, 'seconds')
```

Note that for all other tests in `examples/simple_examples.py`, they all show similar performances between standard tqdm and monitoring thread PR (of course the `no miniters` is slower).

So there is a small performance hit, which is still way better than removing `miniters` completely: removing `miniters` doubles the computation time, whereas the monitoring thread has an overhead of 10% (2s here). The perf hit doesn't even come from the monitoring thread itself, but rather from using `self.miniters` instead of `miniters` in `tqdm.__iter__()`. This seems to slow down by a linear constant, so it's indeed noticeable but it can be ok (or if we can find a way around it would be better but I can't think how). Makes tqdm go from 20s to 22s for the task above.

Note that there is no performance impact on manual tqdm. The perf hit is only on iterable tqdm.

TODO:
- [x] Profile if `self.miniters` in `tqdm.__iter__()` isn't too much of a burden.
- [x] Fix thread hanging up terminal if KeyboardInterrupt.
- [x] Force display refresh on miniters update from monitoring thread (else user will have to wait until next iteration to see the update).
- [x] Fix slowdown at closing (because of `TMonitor.join()`).
- [x] Add unit test for monitoring thread and the special case in #249 (would need to modify `TMonitor` to support virtual discrete timer, just store in `self._time` and `self._sleep` like we do in `tqdm`).
- [x] Add a note to Readme.
- [x] Fix unit test to synchronize with thread and wait for it to wake up instead of sleeping (which makes Travis fail sometimes).
- [x] Fix thread hanging up and making Travis test fail randomly.

### DIFF
--- a/tqdm/tests/tests_tqdm.py
+++ b/tqdm/tests/tests_tqdm.py
@@ -1323,7 +1323,6 @@ def test_monitoring_thread():
     # And a fake tqdm
     class fake_tqdm(object):
         _instances = []
-        pass
 
     # Instanciate the monitor
     monitor = TMonitor(fake_tqdm, maxinterval)
--- a/Makefile
+++ b/Makefile
@@ -48,10 +48,7 @@ all:
 	@+make build
 
 flake8:
-	@+flake8 --max-line-length=80 --count --statistics --exit-zero tqdm/
-	@+flake8 --max-line-length=80 --count --statistics --exit-zero examples/
-	@+flake8 --max-line-length=80 --count --statistics --exit-zero .
-	@+flake8 --max-line-length=80 --count --statistics --exit-zero tqdm/tests/
+	@+flake8 --max-line-length=80 --count --statistics --exit-zero -j 8 --exclude .asv .
 
 test:
 	tox --skip-missing-interpreters
--- a/tox.ini
+++ b/tox.ini
@@ -20,10 +20,7 @@ commands =
 [testenv:flake8]
 deps = flake8
 commands =
-    flake8 --max-line-length=80 --count --statistics --exit-zero tqdm/
-    flake8 --max-line-length=80 --count --statistics --exit-zero examples/
-    flake8 --max-line-length=80 --count --statistics --exit-zero .
-    flake8 --max-line-length=80 --count --statistics --exit-zero tqdm/tests/
+    flake8 --max-line-length=80 --count --statistics --exit-zero -j 8 --exclude .asv .
 
 [testenv:setup.py]
 deps =


lrq3000:
All done, good to go for me. Please review me :)

tqdm/_tqdm.py
```diff
@@ -15,7 +15,9 @@
 from ._utils import _supports_unicode, _environ_cols_wrapper, _range, _unich, \
     _term_move_up, _unicode, WeakSet
 import sys
-from time import time
+from threading import Thread
+from threading import Event
```
CrazyPython: @lrq3000 merge into `from threading import Thread, event`?

tqdm/_tqdm.py
```diff
@@ -41,12 +43,75 @@ def __init__(self, msg, fp_write=None, *a, **k):
...
+        if TMonitor._sleep is not None:
+            self._sleep = TMonitor._sleep
+        else:
+            self._sleep = sleep
+        self.start()
+
+    def exit(self):
+        self.exit_event.set()
+        self.was_killed = True
+        # self.join()  # DO NOT, blocking event, slows down tqdm at closing
```
CrazyPython: @lrq3000 why's this commented code here?

README.rst
```diff
@@ -622,6 +622,43 @@ A reusable canonical example is given below:
...
+operations).
+
+3. configuring ``miniters`` is complicated, ``mininterval`` is much more intuitive,
+so a clever adjustment system ``dynamic_miniters`` was devised: after a few iterations,
+``miniters`` is automatically adjusted to the amount of iterations that equal to
+``mininterval`` time interval. In other words, we try to check if it's time to print
+without checking time at all! Of course, this behavior can be bypassed by setting
+``miniters`` to the value you want.
+
+4. But what if you have fast iterations and then slow iterations? That's the issue with
```
CrazyPython: @lrq3000 inconsistent capitalization, though it's awkward in this situation...

README.rst
```diff
@@ -622,6 +622,43 @@ A reusable canonical example is given below:
...
+``dynamic_miniters``, it will set ``miniters`` to a high number of iterations that will
+be slow to reach with the slow iterations afterwards. A double solution was implemented:
+``maxinterval`` defines the maximum time you should wait before the bar gets a forced
+refresh, and there is a monitoring thread that runs concurrently to ``tqdm`` to check
+if any bar is way over its due refreshing (by comparing to ``maxinterval``), and refresh
+it forcibly.
+
+The monitoring thread should not be a noticeable overhead, and it ensures that
```
CrazyPython: be -> have

README.rst
```diff
@@ -622,6 +622,43 @@ A reusable canonical example is given below:
...
+you will never wait more than 10 seconds by default. This value can be changed
+by setting ``tqdm.tqdm.monitor_interval = x`` where x is the interval (in seconds)
+for the monitor to refresh, and ``maxinterval = x`` for each ``tqdm`` bar.
+
+If you really need to disable any thread for your application, you can set
+``tqdm.tqdm.monitor_interval = 0`` before instanciating any ``tqdm`` bar,
```
CrazyPython: spelling error in "instanciatin"

tqdm/_tqdm.py
```diff
@@ -41,12 +43,75 @@ def __init__(self, msg, fp_write=None, *a, **k):
             super(TqdmDeprecationWarning, self).__init__(msg, *a, **k)
 
 
+class TMonitor(Thread):
+    """ tqdm monitoring thread
```
CrazyPython: the 
```
""" ....
..........
......."""
```

style is consistent, correct?

tqdm/_tqdm.py
```diff
@@ -315,6 +384,11 @@ def _decr_instances(cls, instance):
             for inst in cls._instances:
                 if inst.pos > instance.pos:
                     inst.pos -= 1
+            # Kill monitor if no instances left
```
CrazyPython: if no instances **are** left

tqdm/tests/tests_tqdm.py
```diff
@@ -9,10 +9,14 @@
 import os
 from nose import with_setup
 from nose.plugins.skip import SkipTest
+from nose.tools import assert_raises
```
CrazyPython: code is too nosy

tqdm/tests/tests_tqdm.py
```diff
@@ -9,10 +9,14 @@
...
 
 from tqdm import tqdm
 from tqdm import trange
 from tqdm import TqdmDeprecationWarning
+from tqdm._tqdm import TMonitor
+
+from time import sleep
```
CrazyPython: group builtin imports together - os with time

CrazyPython:
@lrq3000 ding! reviewing done.

tqdm/_tqdm.py
lrq3000: To avoid anyone reimplementing it later. That's useful commented code to avoid regressions.

tqdm/_tqdm.py
lrq3000: No you're right, fixed.

tqdm/_tqdm.py
```diff
@@ -41,12 +43,75 @@ def __init__(self, msg, fp_write=None, *a, **k):
...
+
+
 class tqdm(object):
     """
     Decorate an iterable object, returning an iterator which acts exactly
     like the original iterable, but prints a dynamically updating
     progressbar every time a value is requested.
     """
+
+    monitor_interval = 10  # set to 0 to disable the thread
```
lrq3000: Yes but a few operations every 10 seconds (by default, we can set the monitor_interval higher) has a negligible impact. I profiled and in fact the monitor has no effect that I could notice, it was only `self.miniters` that's producing a 10% overhead (but in exchange we have the guarantee that a tqdm bar never gets frozen, I think it's worth the price!).

Also this change should work with any submodule, that's why I thought it may be fit in the core.

tqdm/tests/tests_tqdm.py
lrq3000: No we (implicitly) adopted the following style advised in Python doc:

```
from __future__ import x

# tqdm internal import
import ._core as tqdm
from ._utils import x

# Third party modules
import x
from x import y  # from are after

# tqdm external imports (from an external module or unit test)
import tqdm
from tqdm import x
```

Ordered alphabetically and with no grouping of modules on one line (no `import os, time, etc.`) but we can group functions/class imports (`from time import sleep, time`).

But that was adopted implicitly so there may be some discrepancies here and there.

lrq3000:
Thank's @CrazyPython, I fixed what you reviewed :)

However I would like a more code oriented review about whether we should include this or not. I think it's a good idea obviously because I'm biased (else I wouldn't have spend the time to implement it), but if you guys can think of some edge case to try this feature out (and see whether it can fail or have shortcomings) it would help a lot!

Or we can merge and see later :) At least it's 100% unit tested, so it doesn't break anything and it's working like expected on the cases it's supposed to fix.

CrazyPython:
@lrq3000 I'll see what I can do... also Travis CI...

lrq3000:
@CrazyPython What about TravisCI?
/EDIT: ah it failed, I don't know why, I restarted the failing tests and it works now without any change. I'm going to try to enhance the unit test, it's because there's a real sleep somewhere to wait for the thread to wake up. I wonder if there's a way around.

CrazyPython:
This may be better as a subclass of tqdm to prevent slowdowns...

tqdm/_tqdm.py
```diff
@@ -716,12 +797,13 @@ def __iter__(self):
                 # Note: does not call self.update(1) for speed optimisation.
                 n += 1
                 # check the counter first (avoid calls to time())
-                if n - last_print_n >= miniters:
+                if n - last_print_n >= self.miniters:
+                    miniters = self.miniters  # watch monitoring thread changes
```
CrazyPython: Can we directly modify `self.miniters` or am I missing something? I'm probably missing something.

tqdm/_tqdm.py
```diff
@@ -315,6 +392,11 @@ def _decr_instances(cls, instance):
             for inst in cls._instances:
                 if inst.pos > instance.pos:
                     inst.pos -= 1
+            # Kill monitor if no instances are left
+            if not cls._instances and cls.monitor:
+                cls.monitor.exit()
+                del cls.monitor
```
CrazyPython: `cls.monitor.exit()` should be called by the `__del__` method of `cls.monitor`.

tqdm/_tqdm.py
CrazyPython: @lrq3000 perhaps it should be more explicit?

tqdm/_tqdm.py
```diff
@@ -41,12 +42,84 @@ def __init__(self, msg, fp_write=None, *a, **k):
...
+        self.was_killed = True
+        # self.join()  # DO NOT, blocking event, slows down tqdm at closing
+        return self.report()
+
+    def run(self):
+        while True:
+            # Sleep some time...
+            self._sleep(self.sleep_interval)
+            # Quit if killed
+            # if self.exit_event.isSet():  # TODO: should work but does not...
```
CrazyPython: hrm, what's this?

tqdm/_tqdm.py
```diff
@@ -41,12 +42,84 @@ def __init__(self, msg, fp_write=None, *a, **k):
...
+            if self.was_killed:
```
CrazyPython: could this be in the while condition?

tqdm/_tqdm.py
```diff
@@ -15,7 +15,8 @@
 from ._utils import _supports_unicode, _environ_cols_wrapper, _range, _unich, \
     _term_move_up, _unicode, WeakSet
 import sys
-from time import time
+from threading import Thread, Event
+from time import time, sleep
 
 
 __author__ = {"github.com/": ["noamraph", "obiwanus", "kmike", "hadim",
```
CrazyPython: dude, credit not given properly...

tqdm/_tqdm.py
```diff
@@ -41,12 +42,84 @@ def __init__(self, msg, fp_write=None, *a, **k):
...
+    sleep_interval  : fload
+        Time to sleep between monitoring checks.
+    """
+
+    # internal vars for unit testing
+    _time = None
+    _sleep = None
+
+    def __init__(self, tqdm_cls, sleep_interval):
+        Thread.__init__(self)
```
CrazyPython: use super call instead?

tqdm/_tqdm.py
CrazyPython: @lrq3000 let's make a bare bones version with #255. Some people would really like it.

lrq3000:
Do not merge this PR, the unit test is not good enough: it's using sleep, but it should synchronize with the thread instead. I tried to synchronize via various ways but the thread randomly hangs up!

That's not good. I hate racing issues and threads... If someone experienced with threads could give a hand, that'd be great.

lrq3000:
Thank's @CrazyPython I will reply to your feedback later, this synchronization issue is critical.

lrq3000:
Replaced `sleep` by an internal wakeup counter check.

However there is still an issue: the thread make the unit test hang sometimes, that's not good. I have no idea what is causing this issue.

lrq3000:
Ok I fixed the hanging thread issue (this was because of a racing condition with the thread waking up too early before tqdm had the time to update, so I had to add additional failsafes). The unit test now runs in a reasonable time.

I did my best to discretize the unit test and make it deterministic, but it's a thread so I cannot guarantee that there won't be any hangup in every possible scenarios. But the whole unit test pass fine so it should be alright to ship for test.

I'll leave it up to you guys whether we release it publicly and remove it if there are negative feedbacks, or wait for other issues to be raised about maxinterval to see if they are all fixed by this PR without issue.

tqdm/_tqdm.py
lrq3000: @casperdcl ordered the credits this way (chronology of arrival for each main contributor). What do you propose as an order?

tqdm/_tqdm.py
lrq3000: No because ```__del__``` cannot be called directly, it's only called by GC. We need a way to signal to the thread that it needs to exit the while loop in `run()`.

tqdm/_tqdm.py
lrq3000: This is an optimization, we store instance variables as local function variables, this speeds up the calculations a great deal because there's no dot reference lookup. But here we need to access self.miniters since it can be modified by the monitoring thread.

tqdm/_tqdm.py
lrq3000: We should use a blocking event to signal that the thread must exit, but it somehow doesn't work so it's using an instance variable instead. This should not change anything in our case, but the best practice would require us to use a blocking event.

tqdm/_tqdm.py
lrq3000: It was before but I removed because it was less reliable. This way we can quit just after going out of sleep, this is a bit more precise and removes some racing conditions.

lrq3000:
Ok I'm done with this PR. Need feedback and/or review + merge. Up to you guys to see.

lrq3000:
@CrazyPython Please remove your comment: https://github.com/lemurheavy/coveralls-public/issues/838#issuecomment-244553672 . They are providing a free service, it's not a professional way to talk to fellow developers.

lrq3000:
@CrazyPython Also they have a life and lots of other issues, 4 days is nothing compared to their workload.

casperdcl:
Nice. Surprised you got a response. I filtered all github emails a few days ago since almost all of them were essentially spam. Pity.

CrazyPython:
@casperdcl ignore those and just use the gh notif system

lrq3000:
@casperdcl Actually we could drop coveralls as we are using codecov's branch coverage (whereas coveralls is only statement coverage as far as I know).

lrq3000:
Merged in commit f78ecc6100c02668704a907a54e880b9c61158d6.

lrq3000:
Self-reminder: next time use (close #id) instead of just (#id) in the commit message, this will automatically define the PR as closed:

http://stackoverflow.com/questions/23015168/how-to-close-a-github-pull-request-with-a-commit-comment

Also [gitreflow](https://github.com/reenhanced/gitreflow) and [git-when-merged](https://github.com/mhagger/git-when-merged) might become handy in the future.

casperdcl:
@lrq3000 did you mean to close this? I intended you to merge #250  into #251, not the other way around. Currently, comparing https://github.com/tqdm/tqdm/pull/250/commits to https://github.com/tqdm/tqdm/pull/251/commits it looks like you're missing things.

lrq3000:
Sorry but too many messages, I processed them one by one yesterday so I didn't see the order before after I was finished with merging this one.

Anyway they are two different PRs, why should they be merged?

lrq3000:
It's normal that the commits history is different, I rebased #250 over #251.

I mean the PR names exactly mean what it's written: #251 was a new feature (the monitoring thread), while #250 was two bugfixes. This is why I didn't merge them, for me it's easier to review and accept bugfixes rather than new features. And it's better to merge them separately in case there are bugs in the future (they will probably be imputable to the new feature #251 rather than the bugfix #250).

casperdcl:
arg ok did you squash merge this branch into master?

lrq3000:
Yes, I thought the review was a signal that I could merge. Also, I did add a few corrections when I merged into master. I will check what is missing since I have a local copy.

lrq3000:
But that's weird, why didn't you want to squash merge? It made more sense to me to squash because it was easier to track down then...

lrq3000:
Ok nevermind, there's just the `interations` typo left in the readme, you merged all the changes, so we'll fix that later :) thank's @casperdcl . But can you tell me what you think about squash merge? We need to be consistent about our merge procedure. I thought that 1 new feature PR = 1 commit. For bugfix PR we might merge or squash merge depending on the case.

casperdcl:
I haven't tagged and formally released master yet so we can squash any minor corrections still. I think minor PR = 1 commit rebase/squash merge. major PR (eg this) = rebase and squash debug commits, then master merge --no-ff.

lrq3000:
Ok I'll do that in the future, thank's for explaining.

2016-10-31 2:58 GMT+01:00 Casper da Costa-Luis <notifications@github.com>:

PR #257, opened 2016-09-02
Faster simple bar for perf test

lrq3000:
Added trick described in https://github.com/tqdm/tqdm/commit/0346a4a19f4cbd7274b1a0bfe12eb54ea71c2319 to the simple bar in perf test. Didn't test if the bar was faster, but since this is exactly the same scenario it should also improve the speed.


lrq3000:
Ok I'm done here. I don't want to enhance this simple_bar further because it should not become a duplicate of `tqdm_bare`, so any enhancement will go exclusively there (except if it's a performance enhancement, but I doubt we can do much more than that).

Issue #255, opened 2016-08-28
Barebones tqdm in submodule


CrazyPython:
add tqdm core for performance freaks. (progress bars often mean performance is needed)

Another related but unrelated note: add extra features using MixIns, so users can create their own tqdm classes with customized features.

lrq3000:
I can add easily and quickly a bare version (please do not call this core, this is reserved to the standard tqdm) using some code we have in perf test, but keep in mind that it will have very few features (no multiple/nested progress bars, no EMA, no bar customization, etc.). It will also be less resilient than standard tqdm (lots of checks that won't be added in bare tqdm).

CrazyPython:
@lrq3000 you can rename other people's issues if you own or have write access to the repo :)

lrq3000:
I know but I preferred to tell you :)

Anyway, I worked on this issue, and in the end such a stripped down tqdm is slower than core tqdm (canonical example: strip dynamic_miniters and you increase the overhead drastically). So a barebone tqdm is still in the works but rather as a minimal example for port developers.

So I'm closing this issue since we likely won't implement it as it was described but repurposed for something else.

About mixins, we'll see later, we already have a modular subclassing system and there's no module yet that could benefit from multiple inheritance of two tqdm submodules.

CrazyPython:
@lrq3000 Hm, what if you just used an old version of tqdm?

lrq3000:
Less features, more buggy and less efficient, hence I dropped this approach. I reused the old approach (function-based with closures) but updated with new features and bugfixes.

CrazyPython:
@lrq3000 Understand. tqdm_bare is slower or faster, and by how much? Also speaking of ports to other languages, why don't you join us on tqdm.cpp?

lrq3000:
It should be faster because there's basically no sanity checking on arguments, less minor features (such as `bar_format`), no class (no object-oriented overhead) and no I/O auto-configuration (console size, environment, etc). So overall fewer opcodes. But it's slower than core tqdm:

```
from tqdm import tbrange
for i in tbrange(int(1e8)):
    pass
100%|#######################|100000000/100000000 [01:20<00:00, 1236089.41it/s]
(80.96664071412025, 'seconds')
for i in tqdm.trange(int(1e8)):
    pass
100%|#######################| 100000000/100000000 [00:19<00:00, 5082075.51it/s]
(19.677953824994802, 'seconds')
```

[...]

PR #258, opened 2016-09-03
Add tqdm_bare and tbrange as minimal example for ports developers

lrq3000:
Implement what was asked in #255.

This PR implements a barebone tqdm with only the basic features. The goal is twofold: maximum speed and a short and easy implementation that other languages can use as a base to implement a minimalist `tqdm`.

I made the choice of keeping it a function with closures. I think this is the best way to maximize performance. But it will impose severe implementation constraints. /EDIT: reimplemented with class-based approach in branch tqdm_bare_class, so we now have both approaches and must choose one!

/EDIT: I realized it's stupid to try to implement a tqdm_bare only for performance reasons, because if we strip out features to the bare, of course the opcode count will drop down significantly but the overhead will rise up: for example, `dynamic_miniters` wouldn't be implemented, which means that by default `miniters=1`. In this case, even if tqdm_bare has fewer opcodes than core tqdm, it would be far slower because it would have a bigger overhead, checking time and printing more often than core tqdm does! And that's normal, core tqdm is already heavily optimized, most features are optional with virtually no cost.

So, since minimizing opcodes is useless, I tried to minimize code statements: the new goals of this module are:
1. Implement major features of `tqdm` in the fastest way possible.
2. Minimize number of code statements.
3. Standalone (should avoid external calls even to `._utils` except to standard lib): the file should contain all logic necessary to reimplement it elsewhere. (I exceptionally call `._utils` to load colorama because I think it's not part of our logic but rather a workaround for Windows console limitations.)

**The final objective is to have a standalone shortened tqdm containing major features as a minimal example for developers working on ports to other languages.**

So here we are: tqdm_bare is 387 lines long including docstrings, comments and spaces and it supports major tqdm features that we love it for, such as automagic nested loops support!

Examples:

```
from tqdm import tqdm_bare, tbrange
from tqdm import tqdm_bare_class, tbcrange
from time import sleep

#######################
# Function based tqdm_bare

for i in tbrange(int(1E6), leave=True, unit='b', unit_scale=True):
    pass

for i in tbrange(10, desc='PAR', leave=True):
    for j in tbrange(100, desc='NESTED', leave=True):
        sleep(0.01)

for i in tbrange(100, desc='ITER', leave=True):
    sleep(0.1)

t = tqdm_bare(total=100, desc='MANU', leave=True)
for i in xrange(100):
    t()
    sleep(0.1)

t = tqdm_bare(desc='NOTOTAL', leave=True)
for i in xrange(100):
    t()
    sleep(0.1)

t = tbrange(10, desc='INIT', leave=True)

##########################
# Class based tqdm_bare_class

for i in tbcrange(int(1E6), leave=True, unit='b', unit_scale=True):
    pass

for i in tbcrange(10, desc='PAR', leave=True):
    for j in tbcrange(100, desc='NESTED', leave=True):
        sleep(0.01)

for i in tbcrange(100, desc='ITER', leave=True):
    sleep(0.1)

t = tqdm_bare_class(total=100, desc='MANU', leave=True)
for i in xrange(100):
    t()
    sleep(0.1)

t = tqdm_bare_class(desc='NOTOTAL', leave=True)
for i in xrange(100):
    t.update()
    sleep(0.1)

t = tbcrange(10, desc='INIT', leave=True)
```

Supported arguments/features (including `dynamic_miniters`):

```
def tqdm_bare(iterable=None, desc=None, total=None, leave=True,
                    file=sys.stdout, width=78, mininterval=0.1,
                    maxinterval=10, miniters=None, disable=False,
                    unit='it', unit_scale=False,
                    smoothing=0.3, initial=0, position=None,
                    **kwargs):
```

Same for `tqdm_bare_class` (and same performance, etc.).

TODO:
- [x] Choose between [tqdm_bare](https://github.com/tqdm/tqdm/blob/tqdm_bare/tqdm/_tqdm_bare.py) (function-based with closures, 387 lines 334 SLOC) and [tqdm_bare_class](https://github.com/tqdm/tqdm/blob/tqdm_bare_class/tqdm/_tqdm_bare.py) (object-oriented, 429 lines 368 SLOC). Or keep both? For reference, core tqdm is 954 lines and 829 SLOC. /EDIT: both are merged in the same branch but in two different scripts `_tqdm_bare.py` and `_tqdm_bare_class.py`. Thus, this will show two different ways of implementing tqdm in other languages, depending on whether the language is functional oriented or object oriented.
- [ ] Unit tests (use the ones for core `_tests_tqdm.py`? Or use old ones from v1.0-2.0?). Goal is to have 100% coverage like core tqdm.
- [ ] Flake8
- [x] Add some missing features that would not impact performance, like `smoothing` (Exponential Moving Average)? (But remember that we don't want to reinvent the wheel, we already have core tqdm).
- [x] Add automated nested loops support
- [ ] Add [new benchmarks](https://github.com/tqdm/tqdm/pull/258#issuecomment-245134732) in examples/simple_examples.py (or create a new `benchmark.py` script?)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,8 @@
+comment:
+  layout: header, changes, diff
+coverage:
+  status:
+    patch:
+      default:
+        target: '80'
+    project: false
--- a/tqdm/__init__.py
+++ b/tqdm/__init__.py
@@ -3,11 +3,14 @@ from ._tqdm import trange
 from ._tqdm_gui import tqdm_gui
 from ._tqdm_gui import tgrange
 from ._tqdm_pandas import tqdm_pandas
+from ._tqdm_bare import tqdm_bare
+from ._tqdm_bare import tbrange
 from ._main import main
 from ._version import __version__  # NOQA
 from ._tqdm import TqdmTypeError, TqdmKeyError, TqdmDeprecationWarning
 
 __all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange', 'tqdm_pandas',
+           'tqdm_bare', 'tbrange',
            'tqdm_notebook', 'tnrange', 'main',
            'TqdmTypeError', 'TqdmKeyError', 'TqdmDeprecationWarning',
            '__version__']
--- a/tqdm/__init__.py
+++ b/tqdm/__init__.py
@@ -2,15 +2,15 @@ from ._tqdm import tqdm
 from ._tqdm import trange
 from ._tqdm_gui import tqdm_gui
 from ._tqdm_gui import tgrange
-from ._tqdm_bare import tqdm_bare
-from ._tqdm_bare import tbrange
+from ._tqdm_bare import tqdm_bare_class
+from ._tqdm_bare import tbcrange
 from ._tqdm_pandas import tqdm_pandas
 from ._main import main
 from ._version import __version__  # NOQA
 from ._tqdm import TqdmTypeError, TqdmKeyError, TqdmDeprecationWarning
 
 __all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange',
-           'tqdm_bare', 'tqdm_brange',
+           'tqdm_bare_class', 'tbcrange',
            'tqdm_pandas',
            'tqdm_notebook', 'tnrange', 'main',
            'TqdmTypeError', 'TqdmKeyError', 'TqdmDeprecationWarning',
--- a/tqdm/__init__.py
+++ b/tqdm/__init__.py
@@ -2,8 +2,8 @@ from ._tqdm import tqdm
 from ._tqdm import trange
 from ._tqdm_gui import tqdm_gui
 from ._tqdm_gui import tgrange
-from ._tqdm_bare import tqdm_bare_class
-from ._tqdm_bare import tbcrange
+from ._tqdm_bare_class import tqdm_bare_class
+from ._tqdm_bare_class import tbcrange
 from ._tqdm_pandas import tqdm_pandas
 from ._tqdm_bare import tqdm_bare
 from ._tqdm_bare import tbrange
--- a/tqdm/__init__.py
+++ b/tqdm/__init__.py
@@ -2,12 +2,16 @@ from ._tqdm import tqdm
 from ._tqdm import trange
 from ._tqdm_gui import tqdm_gui
 from ._tqdm_gui import tgrange
+from ._tqdm_bare import tqdm_bare
+from ._tqdm_bare import tbrange
 from ._tqdm_pandas import tqdm_pandas
 from ._main import main
 from ._version import __version__  # NOQA
 from ._tqdm import TqdmTypeError, TqdmKeyError, TqdmDeprecationWarning
 
-__all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange', 'tqdm_pandas',
+__all__ = ['tqdm', 'tqdm_gui', 'trange', 'tgrange',
+           'tqdm_bare', 'tqdm_brange',
+           'tqdm_pandas',
            'tqdm_notebook', 'tnrange', 'main',
            'TqdmTypeError', 'TqdmKeyError', 'TqdmDeprecationWarning',
            '__version__']
--- a/tqdm/_tqdm_bare.py
+++ b/tqdm/_tqdm_bare.py
@@ -157,13 +157,17 @@ def tqdm_bare(iterable=None, desc=None, total=None, leave=True,
                     if maxinterval and delta_t >= maxinterval and mininterval:
                         # Set miniters to correspond to maxinterval
                         miniters[0] = delta_it * mininterval / delta_t
-                    else:
+                    elif smoothing:
                         # EMA-weight miniters to converge
                         # towards the timeframe of mininterval
-                        temporalterm = (mininterval / delta_t
-                                            if mininterval and delta_t else 1)
-                        miniters[0] = smoothing * delta_it * temporalterm + \
-                                   (1 - smoothing) * miniters[0]
+                        miniters[0] = smoothing * delta_it * \
+                                            (mininterval / delta_t
+                                             if mininterval and delta_t
+                                             else 1) + \
+                                            (1 - smoothing) * miniters[0]
+                    else:
+                        # Maximum nb of iterations between 2 prints
+                        miniters[0] = max(miniters[0], delta_it)
 
                 # EMA rate
                 if smoothing and delta_t:
@@ -248,13 +252,17 @@ def tqdm_bare(iterable=None, desc=None, total=None, leave=True,
                         if local_maxinterval and delta_t >= local_maxinterval and local_mininterval:
                             # Set miniters to correspond to maxinterval
                             local_miniters = delta_it * local_mininterval / delta_t
-                        else:
+                        elif smoothing:
                             # EMA-weight miniters to converge
                             # towards the timeframe of mininterval
-                            temporalterm = (local_mininterval / delta_t
-                                                if local_mininterval and delta_t else 1)
-                            local_miniters = local_smoothing * delta_it * temporalterm + \
-                                       (1 - local_smoothing) * local_miniters
+                            local_miniters = local_smoothing * delta_it * \
+                                                 (local_mininterval / delta_t
+                                                 if local_mininterval and delta_t
+                                                 else 1) + \
+                                                (1 - local_smoothing) * local_miniters
+                        else:
+                            # Maximum nb of iterations between 2 prints
+                            local_miniters = max(local_miniters, delta_it)
 
                     # EMA rate
                     if local_smoothing and delta_t:
--- a/tqdm/_tqdm_bare.py
+++ b/tqdm/_tqdm_bare.py
@@ -166,7 +166,7 @@ class tqdm_bare(object):
         self.fp.write(_unicode('\n' * pos + _term_move_up * -pos))
     
     def __init__(self, iterable=None, desc=None, total=None, leave=True,
-                    file=sys.stdout, width=78, mininterval=0.1,
+                    file=None, width=78, mininterval=0.1,
                     maxinterval=10, miniters=None, disable=False,
                     unit='it', unit_scale=False,
                     smoothing=0.3, initial=0, position=None,
@@ -177,6 +177,11 @@ class tqdm_bare(object):
             self.iterable = iterable
             self.disable = disable
             return
+
+        # Define file default value at instanciation rather than class import
+        # (ease file redirection)
+        if file is None:
+            file = sys.stdout
 
         # Preprocess the arguments
         if total is None and iterable is not None:
--- a/tqdm/_tqdm_bare.py
+++ b/tqdm/_tqdm_bare.py
@@ -1,0 +1,428 @@
+"""
+Standalone minimal tqdm but with major features.
+For developers porting to other languages or speed freaks.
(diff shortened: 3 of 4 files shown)


lrq3000:
I stop here, I'll add unit tests if enough people are interested.

lrq3000:
I did some benchmark against different versions of tqdm:

```
bare: 100%|#################|100000000/100000000 [00:20<00:00, 5079080.71it/s]
v1.0: |##########| 100000000/100000000 100% [elapsed: 00:19 left: 00:00, 5175179.85 iters/sec]
noclass: 100%|##########| 100000000/100000000 [00:19<00:00, 5159426.28 it/s]
class: 100%|##########| 100000000/100000000 [00:20<00:00, 4893086.06it/s]
core: 100%|#################| 100000000/100000000 [00:19<00:00, 5140595.26it/s]
```

* Bare is this PR
* v1.0 is the [original tqdm](https://github.com/noamraph/tqdm)
* [noclass](https://github.com/tqdm/tqdm/commit/39e814b2d692479b677aef46618d2881fb8cb791) is tqdm v1.0 with optimizations (status_printer is not a class anymore but a function with closures), just before passing to class-based approach in v2.0.
* [class (tqdm v2.0)](https://github.com/tqdm/tqdm/commit/35773af7473c9ae1927024d599b32874cb4b1b7c) is when we transitioned original tqdm to a class based approach instead of function based with closures. I took a commit where the class based approach was fixed to avoid the slowdown (by inlining the same code in `__iter__()` and `update()`).
* [core is current core tqdm](https://github.com/tqdm/tqdm/commit/02cbd9c2f527ab1dce28dd3f155650f233c28519).

All bars were benchmarked with `trange(int(1e8), miniters=4500000, mininterval=0.1, desc='bare', leave='True')`. Note that without `miniters=4500000`, only core tqdm and bare tqdm achieve reasonable performance at `4500000 it/s` whereas all others drop significantly.

So as we can see, tqdm_bare is somewhat as fast as all the other bars, except class (tqdm v2.0) which is significantly slower than tqdm_bare.

Note that core tqdm is still sensibly faster than tqdm bare, which shows once again that we did a pretty good job at optimizing core tqdm. But I don't know why this is so, since there is technically less opcodes, so my guess is that the function based approach with closures is slower than a class based approach when we implement some of the major features of tqdm.

I'm going to try to reimplement tqdm_bare with a class based approach and see if it speeds up a bit.

CrazyPython:
@lrq3000 so `noclass` is faster than `v1.0`?

casperdcl:
v1.0 was slooooooow.

CrazyPython:
@casperdcl I mean slower

CrazyPython:
For comparison:
```
bare	5079080.71
v1.0	5175179.85
noclass	5159426.28
class	4893086.06
core	5140595.26
```

lrq3000:
I pushed a class-based version of tqdm_bare the [branch tqdm_bare_class](https://github.com/tqdm/tqdm/tree/tqdm_bare_class), and did a few additional optimizations.

Here are the results (manually averaged over 10 runs):

```
bare: 100%|#################|100000000/100000000 [00:18<00:00, 5312367.22it/s]
bare_class: 100%|#################|100000000/100000000 [00:18<00:00, 5285691.66it/s]
v1.0: |##########| 100000000/100000000 100% [elapsed: 00:18 left: 00:00, 5366246.31 iters/sec]
noclass: 100%|##########| 100000000/100000000 [00:18<00:00, 5368839.28 it/s]
class: 100%|##########| 100000000/100000000 [00:20<00:00, 4980823.85it/s]
core: 100%|#################| 100000000/100000000 [00:19<00:00, 5223841.64it/s]
```

It seems the standard deviation is ~100K it/s.

So what we can conclude:
* class tqdm (v2.0) is the slowest, because it didn't have yet all the optimizations that we have done in core tqdm up to now.
* original tqdm v1.0 and noclass (v1.0 with optimizations and a few more features) are about the same speed.
* core tqdm is a bit slower than original tqdm, but since it contains a LOT more features that's really quite a feat.
* bare and bare_class are about the same speed, and their speed is between core tqdm and original tqdm. That's very good considering it has most of core tqdm features. But the fact that there's virtually no difference between the two means that a function based approach does not alleviate the overhead of other features. Note also that I didn't optimize them as heavily as core tqdm, so we can probably gain a bit more speed at the expense of readability (which I don't want given the new goal).

So I stop here and leave it up to you guys to choose whether we implement `tqdm_bare` (function-based approach, 386 lines 333 SLOC) or `tqdm_bare_class` (object-oriented approach, 428 lines 367 SLOC).

Please feedback!

lrq3000:
For reference, core tqdm is currently 954 lines and 829 SLOC, so this is a drastic reduction of SLOC while keeping most features. This should greatly ease the porting to other languages, as an implementation of tqdm_bare can easily be extended to support everything from core tqdm.

CrazyPython:
wow. applause to use sir.

lrq3000:
Ah sorry guys I didn't see your replies before posting my latest comments.

@CrazyPython Thank you, that was more work than I intended to put on this issue :) However I'm quite surprised that with fewer SLOC this PR isn't faster than tqdm v1.0 considering the optimizations we have done, but I guess that the new features have more overhead than the gain we had from optimizations. But they are worth the small drop if you ask me (nested bars are so useful!).

About the rate numbers, @casperdcl is right: v1.0 is the slowest, except if you set manually a miniters. If you don't, v1.0 has no dynamic_miniters, so the rate is ~2150000, half the speed of any other tqdm version I used in the benchmark. So I specified a good miniters value computed by dynamic_miniters in core tqdm as a basis for all the benchmarks.

So we can say that the perf you see for v1.0 is a bit of a cheat because you will never guess this miniters value by yourself in a real application, but that's still good as a higher bound for the performance a progress bar can achieve (if everything is absolutely perfect!).

Also the rate isn't absolute, the numbers vary greatly from run to run because of CPU handling by OS (I try to stop most other processes but some are always in the background), so they usually vary by 50K to 100K.

If I would rate the range of perf per tqdm version, I would say:

```original = noclass > bare = bare_class > core >> class (by far the slowest, big gap with core tqdm)```

**Comment by lrq3000** [comment]
Here are the results without miniters tuning:

```
bare	2123341
v1.0	2075808
noclass	5175179
class	4942420
core	4427913
```

Clearly there's a bug with bare calculation of dynamic_miniters. I'm going to fix that.

But still, here's the ranking:
```noclass > class >> core >> bare > v1.0```

**Comment by lrq3000** [comment]
Ah no bug in fact, it's because I set smoothing=0 only for tqdm_bare (I can't remember why) in my benchmark, which disables dynamic_miniters. The same issue happens with core tqdm. Maybe we should fix that? (special calculation with smoothing = 0?).

New timing for bare:
```
bare    4288357
```

**Comment by lrq3000** [comment]
It's quite interesting to note that while `class` was the slowest with fixed miniters, it's the second fastest with dynamic_miniters.

**Comment by CrazyPython** [comment]
so how do get he fastest barest bar, with or without miniters?

**Comment by lrq3000** [comment]
@CrazyPython Fastest bar is original tqdm v1.0 with tuned miniters (because we basically skip time check, I think that setting mininterval=0 would give just the same result), with noclass being second. Without tuned miniters, noclass is the fastest. So globally, noclass is the fastest of them all.

**Comment by lrq3000** [comment]
Ah! Interesting finding, while I was fixing the dynamic_miniters not working with smoothing=0 bug (I will merge this fix with #250), I found that the old way of calculating dynamic_miniters speeds up drastically `tqdm`!

```
core    5241090
```

So with this fix, we have the following:

```
bare    4488357
v1.0    2075808
noclass 5175179
class   4942420
core    5241090
```

I can probably apply the same fix to bare and get the same boost.

So we now have the following ranking:

`core > noclass > class > bare >> v1.0`

Interesting!

And if you are wondering why, the reason is that we are simply doing `self.miniters = max(self.miniters, delta_it)` without any EMA, so `tqdm` is converging faster to the optimal miniters for this particular case where every iteration is blazingly fast and constant. With a jittery iterable, I would expect the EMA calculation to better fit (faster tqdm because faster adaptation to changes).

**Comment by CrazyPython** [comment]
Much progress on speed in general.

**Comment by lrq3000** [comment]
Here are the final results of the benchmarking and optimizing, everything is clarified:

* So I applied the same dynamic_miniters fix, and we get the same boost for tqdm_bare, so now with `miniters=None` we have `bare = noclass = core` which are both the fastest (but need to set `smoothing=0` for bare and core, this reproduces the calculation in noclass).

* For the fixed miniters case, remember that `v1.0 = noclass > bare > core`. I found out that the culprit is ```__init__.py```. Yes you heard right: if we strip everything but `tqdm_bare` in ```__init__.py```, we get `noclass = bare`.

How did I find out about ```__init__.py``` impact on performance? Because I made a copy of `noclass` and reimplemented code blocks from `bare` little by little, and I found absolutely no difference in performance. Then I tried to put it as a module in current tqdm, instead of direct access to `tqdm.py` like in the past, and this dropped the performance.

This is an additional argument to implement the new submodules architecture of #198.

So in the end, `tqdm_bare` now has the best performance all around like `noclass` but with the latest major features (like nested progress bars). I think that's good enough.

But the benchmarking was useful since it allowed to spot two performance regression: no dynamic_miniters calculation when smoothing=0 (and speed impact) + ```__init__.py``` impact on performance.

**Comment by lrq3000** [comment]
Also note that to benchmark correctly, you need to restart twice your test script: the first run will recompile `.pyc` files, and the second run will use them. You can KeyboardInterrupt the first run as soon as the bar you are interested to test is printed.

I used this trick extensively to test ```__init__.py```, because else the first run would always be wrong and use the precedent bytecode compilation.

**Comment by CrazyPython** [comment]
@lrq3000 benchmarks should have a wrapper script or this needs to be documented somewhere less obscure than here.

**Comment by CrazyPython** [comment]
`tqdm_bare_class` - it's supposed to be a guide for port developers

**Comment by lrq3000** [comment]
@CrazyPython This scenario could be added to `examples/simple_examples.py`:

```
# Fixed miniters benchmark
from tqdm import tbrange
for i in tbrange(int(1e8), miniters=None, mininterval=0.1, desc='bare', leave='True', smoothing=0):
    pass

# Dynamic miniters benchmark
from tqdm import tbrange
for i in tbrange(int(1e8), miniters=4500000, mininterval=0.1, desc='bare', leave='True', smoothing=0):
    pass
```

About `tqdm_bare` vs `tqdm_bare_class`, I just benchmarked both after all of the latest changes, and they basically have the same performance.

Both are interesting: I think `tqdm_bare_class` is a bit nicer even if a bit more verbose (because object-oriented constructs are verbose), but `tqdm_bare` is interesting because it shows how `tqdm` can be implemented in a functional only language. Could be interesting for functional languages like OCAML?

Maybe we should keep both?

CrazyPython:
@lrq3000 ooh, both is a nice idea. Let's do that!

lrq3000:
I also now think we should keep both, because the new goal of this submodule is to provide a simplified version for porting to other languages, and non-object-oriented languages will benefit from `tqdm_bare` whereas object oriented languages can implement `tqdm_bare_class`.

What do you think about it @casperdcl ? Should we keep both?

lrq3000:
I merged both in the same branch, it will be easier to review, compare and test.

casperdcl:
oh yes, nice. I'd agree that we should have both a class and a functional approach for future ref. Probably would merge this in after the next major release since it doesn't actually add features though.

lrq3000:
Ok great :) yes every new submodules should be merged after the new
submodule architecture (which is next major release).
Le 12 Nov. 2016 17:56, "Casper da Costa-Luis" <notifications@github.com> a
écrit :

PR #263, opened 2016-09-11
Disable coveralls because of duplicate comments on github

lrq3000:
Coveralls is currently spamming our PR everytime we commit with duplicate comments containing the exact same info. This seems to have started when we merged #232, more exactly when we started caching pip packages like pandas. The error was reported by @CrazyPython to coveralls devs (https://github.com/lemurheavy/coveralls-public/issues/838) but it was closed and their fix doesn't seem to work.

Plus it reduces build time (~5 sec per python version) plus coveralls sometimes fails and makes the build fail.

Anyway, we use codecov since a while as our main coverage report because it is more stringent (branch coverage for codecov instead of statement coverage for coveralls). So we can safely disable it without losing anything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,15 +42,16 @@ install:
   # https://docs.travis-ci.com/user/caching/#Clearing-Caches
   - pip install pandas
   # Coverage install
-  - pip install tox 'coverage<4' coveralls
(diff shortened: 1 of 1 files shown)


lrq3000:
@casperdcl You should now be able to reactivate your email notifications without receiving spam comments from coveralls.

Issue #234, opened 2016-08-11
tqdm_auto to auto detect submodule to use depending on frontend


lrq3000:
Maybe we can make a `tqdm_auto` submodule which would try to autodetect the frontend env and provide the most appropriate module, eg:
- tqdm_notebook for [IPython notebook](http://stackoverflow.com/questions/15411967/how-can-i-check-if-code-is-executed-in-the-ipython-notebook).
- tqdm_gui for IDEs not supporting `\r` control character such as [IDLE](http://stackoverflow.com/questions/17133769/find-out-if-a-python-script-is-running-in-idle-or-terminal-command-prompt/17135271#17135271) and PyCharm.
- core tqdm for the rest.

The proposition is interesting because it would fix three issues:
- Some users are targetting multiple envs, currently they have to fix this by themselves.
- Some IDEs do not support the required control characters, so there's no way to fix core tqdm display issues. But by providing a `tqdm_auto` wrapper, users could just use that in most instances and get a correctly displayed progress bar.
- With tqdm being more and more included in various libraries, the risk that these libraries get used in various development environment increases and thus the usage of tqdm. In these instances, tqdm should be resilient enough to ensure that it's not the breaking point for a library (it must not prevent the usage of a library in any env).

What do you think about it guys?

thehesiod:
it seems in PyCharm 2016.3.2 using one progressbar works ok, but if you add multiple progress bars it breaks down.  Not sure if this is a bug in PyCharm's terminal or tqdm

lrq3000:
See #372 for auto detection between commandline and notebook (thanks to @flying-sheep!).

lrq3000:
@thehesiod yes that's a known issue of IDLE/PyCharm, we can't do anything about it. Hence why this feature request, it will open a GUI to handle multiple bars in IDLE/PyCharm.

flying-sheep:
another alternative would be to horizontally tile progress bars, so one `\r` still clears everything, or using pager mode in case those limited terminal emulators support clearing the screen or positioning the cursor there.

lrq3000:
Yes but then we would quickly run out of space and the bars will be forced
to have a newline return, so with this technique we would only have max 2/3
simultaneous bars :-/

2017-04-11 15:58 GMT+02:00 Philipp A. <notifications@github.com>:

mlubej:
Hello!

Any news on this? :D

flying-sheep:
[...]

PR #267, opened 2016-09-13
cleaner tqdm_notebook exit when not run in a notebook

to266:
When not run in a notebook, the `sp` is not set. This checks if it is set, and exits cleanly.

The idea is to be able to do something like this (with appropriate catching if interested..):

```python
from tqdm import tqdm, tqdm_notebook

def progressbar(*args, **kwargs):
    try:
        return tqdm_notebook(*args, **kwargs)
    except:
        return tqdm(*args, **kwargs)
```

This is a temporary workaround I found useful while waiting for https://github.com/tqdm/tqdm/issues/234

### DIFF
--- a/tqdm/_tqdm_notebook.py
+++ b/tqdm/_tqdm_notebook.py
@@ -203,15 +203,17 @@ class tqdm_notebook(tqdm):
 
     def close(self, *args, **kwargs):
         super(tqdm_notebook, self).close(*args, **kwargs)
-        # Try to detect if there was an error or KeyboardInterrupt
-        # in manual mode: if n < total, things probably got wrong
-        if self.total and self.n < self.total:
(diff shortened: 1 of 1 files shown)


CrazyPython:
@to266 fix the build failure

CrazyPython:
@lrq3000 reset the build, travis made an api error again

lrq3000:
Damn coveralls, it should be disabled by now #263.

CrazyPython:
@lrq3000 not spamming anymore though...

> Damn coveralls, it should be disabled by now #263.

This is what I love about open source and software development in general. Our non-for-profits are lean and snarky startups. And being able to say this too.

lrq3000:
Thank you @to266 for the clean PR, merging it right now.

Issue #266, opened 2016-09-13
Support custom metrics in progress bar format


ferrouswheel:
It'd be nice to update custom metrics on the progress bar status. This would make tqdm super helpful for something like machine learning training where you want to know how well your model is doing.

e,g. before:

```
0%|3                                             | 27/6000 [08:24<28:42:16, 17.30s/it]
```

after:

```
0%|3                                | 27/6000 [08:24<28:42:16, 17.30s/it, 0.1233 loss]
```

lrq3000:
Hello @ferrouswheel,

You can supply additional information in the description:

```
from tqdm import trange
t = trange(6000, desc='ML')
for i in trange:
    loss = compute_loss() # your function here
    t.set_description('ML (loss=%g)' % loss)
```

We could add a `set_postfix()` function to allow you to add information at the end of the bar, but this would be a slight additional overhead.

lrq3000:
PS: please look the previous comment on github, not in the email, I fixed an error in the code.

CrazyPython:
@lrq3000 this would be at the left though - maybe add a `set_postfix()` function in a submodule? Or a simple subclass recipe.

ferrouswheel:
@lrq3000 Thanks - that's awesome :+1: 

I think I assumed it should go on the right because unless the loss text width is fixed it'll make the progress bar jump around a bit, but that's easy enough to avoid.

The implementation I thought might be most powerful is being able to provide tqdm a template "context", which the progress bar format string could reference variables from. Not sure what the performance implications are though....

lrq3000:
@ferrouswheel Please see if #270 meets your needs better than `set_description()`.

ferrouswheel:
Thanks for implementing this!

PR #270, opened 2016-09-15
Add set_postfix

lrq3000:
Add a `postfix` argument and `set_postfix()` method to supply additional stats to display on the bar, as suggested in #266. The method automatically formats the inputs depending on their datatype. This could be useful for machine learning and for data scientists.

This is an alternative to `set_description()` which asks the user to format themselves the whole description.

Example:

```
from tqdm import trange
from random import random, randint
from time import sleep

t = trange(100)
for i in t:
    t.set_postfix(loss=random(), gen=randint(1,999), str='h', lst=[1, 2])
    sleep(0.1)
```

Output:

```
 98%|9| 98/100 [00:09<00:00,  9.90it/s, loss=0.0419, lst=[1, 2], gen=677, str=h]
```

TODO:
- [x] Test perf using examples/simple_examples.py to see if impact negligible or significative when not using postfix (because of `self.postfix` access in `__iter__()`, just like `self.desc`) -> no difference, the impact is negligible.
- [x] Add unit tests
- [x] Fix flake8 warning
- [x] Add to readme
- [x] Add support in tqdm_notebook and potentially other submodules (already works, no change needed, as long as `r_bar` is managed, the postfix is also handled implicitly).
- [x] Support for ordered arguments.
- [x] Rebased.

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -717,7 +717,8 @@ class tqdm(object):
         self._time = time
         self.bar_format = bar_format
         self.postfix = None
-        if postfix: self.set_postfix(**postfix)
+        if postfix:
+            self.set_postfix(**postfix)
 
         # Init the iterations counters
         self.last_print_n = initial
@@ -1063,8 +1064,10 @@ Please use `tqdm_gui(...)` instead of `tqdm(..., gui=True)`
             # Else for any other type, try to get the string conversion
             elif not isinstance(kwargs[key], basestring):
                 kwargs[key] = str(kwargs[key])
+            # Else if it's a string, don't need to preprocess anything
         # Stitch together to get the final postfix
-        self.postfix = ', '.join('%s=%s' % (key, kwargs[key]) for key in kwargs.keys())
+        self.postfix = ', '.join('%s=%s' % (key, kwargs[key])
+                                 for key in kwargs.keys())
 
     def moveto(self, n):
         self.fp.write(_unicode('\n' * n + _term_move_up() * -n))
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -15,6 +15,7 @@ from __future__ import division
 from ._utils import _supports_unicode, _environ_cols_wrapper, _range, _unich, \
     _term_move_up, _unicode, WeakSet
 import sys
+from numbers import Number
 from threading import Thread
 from time import time
 from time import sleep
@@ -199,7 +200,7 @@ class tqdm(object):
     @staticmethod
     def format_meter(n, total, elapsed, ncols=None, prefix='',
                      ascii=False, unit='it', unit_scale=False, rate=None,
-                     bar_format=None):
+                     bar_format=None, postfix=None):
         """
         Return a string-based progress bar given some parameters
 
@@ -240,6 +241,8 @@ class tqdm(object):
             '| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt}]'
             Possible vars: bar, n, n_fmt, total, total_fmt, percentage,
             rate, rate_fmt, elapsed, remaining, l_bar, r_bar, desc.
+        postfix  : str, optional
+            Same as prefix but will be placed at the end as additional stats.
 
         Returns
         -------
@@ -284,8 +287,9 @@ class tqdm(object):
             # format the stats displayed to the left and right sides of the bar
             l_bar = (prefix if prefix else '') + \
                 '{0:3.0f}%|'.format(percentage)
-            r_bar = '| {0}/{1} [{2}<{3}, {4}]'.format(
-                    n_fmt, total_fmt, elapsed_str, remaining_str, rate_fmt)
+            r_bar = '| {0}/{1} [{2}<{3}, {4}{5}]'.format(
+                    n_fmt, total_fmt, elapsed_str, remaining_str, rate_fmt,
+                    ', '+postfix if postfix else '')
 
             if ncols == 0:
                 return l_bar[:-1] + r_bar[1:]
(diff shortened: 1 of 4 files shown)


ferrouswheel:
This looks great - can't wait to use it :+1:

tqdm/_tqdm.py
```diff
@@ -944,8 +945,10 @@ def set_postfix(self, **kwargs):
             # Else for any other type, try to get the string conversion
             elif not isinstance(kwargs[key], basestring):
                 kwargs[key] = str(kwargs[key])
+            # Else if it's a string, don't need to preprocess anything
```
CrazyPython: @lrq3000 Should this be indented like this or moved one line before the code?

lrq3000: This is intended like this to explicitly explain that we don't preprocess strings. This comment can be removed as it can be deduced by other conditions, but I like to be explicit.

lrq3000: I call this the "ghost else" (I made it up): it often happens that else conditions are not explicitly set but implicitly, but the other conditions are designed with this implicit "ghost else" in mind. So in these cases, I add a comment to explain what the "ghost else" is meant to do.

CrazyPython: @lrq3000 yes, but should it be moved to before... wait the code changed... but it didn't... I don't know what drug I was on when I reviewed that. Damn.

casperdcl:
any reason this logic can't be made part of prefix? we really need to think about adding new kwargs - the current list is long enough to be a mswin32 kernel call.

lrq3000:
@casperdcl Haha you're right, the list is getting quite long. The init kwarg is not necessary but it fits with our standard to provide arguments to set the bar display at init. So if we remove postfix at init, we could also remove desc as it can be set through set_description.

Anyway, about the core of question, the logic of postfix is different: it's aimed to provide additional statistics that can change a lot from one display to the next. This implies two things:

1. if it was supplied in desc, then this would change/move the bar size from one display to the next, which is not pretty to say the least.
2. if it was supplied in desc, the user would need to supply one's own formatting template, like `set_description('cost: %f, state: %s, etc.' % (cost, state))`. With a separate postfix method, we get to provide a more transparent way to supply additional stats without requiring the user to provide the formatting template, tqdm does it automagically depending on arguments datatypes: `set_postfix(cost, state)`. This is a killer feature for machine learning afficionados.
3. people want to see the additional stats on the right side of the bar, along the iteration rate and time, because it makes more sense to them...

So I thought that postfix was nice to have since it's not that complex and it doesn't add any overhead.

casperdcl:
how about this?
- rename `set_description(s)` to `set_description(*a, **k)`
- add: `if len(a) > 1 or k: set_postfix(*a, **k)`

lrq3000:
Then you cannot have a description and a postfix at the same time. But at
the root the difference is semantic: set a description on the left, or add
additional infos on the right.

2016-11-12 16:09 GMT+01:00 Casper da Costa-Luis <notifications@github.com>:

casperdcl:
hmm *args as desc and **kwargs as postfix?

lrq3000:
@casperdcl Yes functionally it would be about the same but the API will probably be a bit more confusing (people will have to understand that *args is for the left size, whereas **kwargs is for the right side). Also it would impact the performance of set_description() which does not contain any conditionals currently, but will if we merge it with set_postfix().

Soares:
+1 to this feature. FWIW, as a user of tqdm, I think it's worth having a "postfix" kwarg in the constructor that is separate from "description"; it does seem to me that they are serving different purposes. Also, as someone who does machine learning work, I would much prefer first-class postfix support that lets me say "here are a bunch of metrics for my model; add them to the progress bar please" without having to think about weird ways that that interacts with `description` in the name of trying to combine kwargs that are actually there for different reasons.

Also, I'll note that for my use case, it's valuable to be able to affect the order of the postfix (using, e.g., a python ordereddict). This is a pretty trivial change to add to this PR, I'd write it something like this:

```
    def set_postfix(self, ordered_dict=None, **kwargs):
        """
        Set/modify postfix (additional stats)
        with automatic formatting based on datatype.
        """
        from collections import OrderedDict
        postfix = OrderedDict([] if ordered_dict is None else ordered_dict)
        for key in sorted(kwargs.keys()):
            postfix[key] = kwargs[key]
        # Preprocess stats according to datatype
        for key in postfix.keys():
            # Number: limit the length of the string
            if isinstance(postfix[key], Number):
                postfix[key] = '{0:2.3g}'.format(postfix[key])
            # Else for any other type, try to get the string conversion
            elif not isinstance(postfix[key], _basestring):
                postfix[key] = str(postfix[key])
            # Else if it's a string, don't need to preprocess anything
        # Stitch together to get the final postfix
        self.postfix = ', '.join('%s=%s' % (key, postfix[key])
                                 for key in postfix.keys())
```

Also, I'll note that while a "x=y, a=b, c=d" type postfix works for my purposes, I'd lean towards a slightly more general machinery that lets the user put in their own format string if they want to.

lrq3000:
Thanks for the suggestion, I didn't think an order in parameters really
mattered as long as they were displayed, but this does not really cost much
so this can be added :) Thank you for the code sir!

About the custom formatting, yes you can by providing strings, their
formatting is always respected:

`tqdm.set_postfix(mystat='{0:1.5f}'.format(mystat))`

This was the most pythonic way I could think to provide both custom
formatting and easy automatic formatting at the same time (ie, if the type
of an argument is not a string, then the auto formatting kicks in), but if
you have another suggestion feel free :)

2016-12-19 21:34 GMT+01:00 Nate <notifications@github.com>:

lrq3000:
All good, ready to merge! This PR is strategic as it will appeal to a lot of machine learning afficionados. Combined with `tqdm_notebook`, `tqdm` has the potential to become the defacto library for machine learning notebooks!

casperdcl:
Ok you convinced me. Happy machine learning, folks!

lrq3000:
Thanks Casper! :D

Issue #265, opened 2016-09-13
Customised step size per iteration


ferrouswheel:
When I first used tqdm with `tqdm(range(0,1000,10))` I was half expecting for it to automatically detect that the iterator increases by 10 every iteration, counting from 0 to 1000. Instead it counts to 100.

While it'd be great to infer this from the wrapped iterator, being able to give a hint about the start/end and step size would also work.

CrazyPython:
This functionality already exists.

ferrouswheel:
Ah sorry. I couldn't find it by examples, documentation, or a quick look at
the code. Care to point me in the right direction?

CrazyPython:
@ferrouswheel use the source! it's a fairly obvious keyword variable.

CrazyPython:
manual `.update()` is also supported:

```python
import tqdm
bar = tqdm.tqdm(total=somenumber)
for blah in blahblah:
    bar.update(x)  # x = step amount
```

CrazyPython:
@ferrouswheel `total` should be it

ferrouswheel:
Thanks! I think my confusion is in the wording, "number of expected iterations". Since my iteration count (number of loops) would be different to the total value that tqdm counts up to. I also didn't understanding that I could update tqdm directly with an arbitary value, but that's probably me reading the docs a little too quickly.

lrq3000:
In fact, `trange` bases its automatic calculation of `total` based on what `len()` is returned by `range()`. In your case, the `len`, or the number of loop iterations, is indeed 100 because you have a stepsize of 10 for 1000.

In `trange()`, we could use what is returned by the iterator to update `self.n`, but we still won't be able to know what is the original total since it is not returned by `range()` (only the concrete number of loop iterations is returned by `len(range(0, 1000, 10))`, which is 100, not the total 1000).

ferrouswheel:
@lrq3000 I thought maybe a keyword argument to `trange` might be a nice way to tell it the `total`, and then the update step would be `total/len(range(..)`.

CrazyPython:
@lrq3000 well, you could just parse the arguments passed to trange... I mean they're right there.

lrq3000:
Nope we don't have neither the real total neither the stepsize, that's what
I've been trying to explain. But Joel's suggestion might be feasible, I'll
have a look.

2016-09-13 22:55 GMT+02:00 CrazyPython <notifications@github.com>:

CrazyPython:
[...]

PR #273, opened 2016-09-17
Add stepsize + autodetection of initial, total and stepsize in trange()

lrq3000:
Implement #265 with 2 strategies: autodetection of stepsize if `len(iterable) != total` (total is explicitly specified) + autodetection in `trange()` of `range()` arguments to feed `tqdm()` (using a new function `trange_preprocess()` -- I wanted to turn `trange()` into a class but then the wrapper did not work anymore, it cannot just return a wrapped range under tqdm without wrapping in trange also).

TODO:
- [ ] Call `trange_preprocess()` in all `trange()`-like of submodules.
- [x] Test impact on performance (`examples/simple_tests.py`) -> significative difference: 15% in manual mode and 5-10% in iterable mode. All because of `tqdm.stepsize`. If we remove `tqdm.stepsize` but keep `trange_preprocess()`, we would remove all performance impact, but iterable mode tqdm won't be able to account for `range()`'s stepsize (only `initial` and `total`, which is already good but not perfect). What should we do @casperdcl ?
- [ ] Add unit test

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -562,11 +562,11 @@ class tqdm(object):
         elif total is not None and iterable is not None and stepsize is None:
             try:
                 iterable_total = len(iterable)
-                stepsize = max(int(iterable / iterable_total), 1)
+                stepsize = max(int(total / iterable_total), 1)
             except (TypeError, AttributeError):
                 pass
 
-        if stepsize is None:
+        if stepsize is None or stepsize <= 0:
             stepsize = 1
 
         if ((ncols is None) and (file in (sys.stderr, sys.stdout))) or \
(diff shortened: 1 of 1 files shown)


casperdcl:
hmm, is this essentially a scale factor? eg instead of displaying 123/1000, show 234/2000?

lrq3000:
For stepsize yes, this is essentially a scale factor as far as i can see.
Le 12 Nov. 2016 17:29, "Casper da Costa-Luis" <notifications@github.com> a
écrit :

casperdcl:
In that case I'd rather have this implemented as a `unit_scale=x` and passed directly to the formatter, where `x=0|False|1|True` are treated as before and other values multiply `total` and `n`.

lrq3000:
Yes and they should multply at printing, like the other PR proposed. Indeed
this would probably reduce the overhead a lot, because I was not happy with
the way I implemented it, it was too costly. That's a good idea Casper :)

2016-11-13 11:56 GMT+01:00 Casper da Costa-Luis <notifications@github.com>:

casperdcl:
closing in favour of #378

Issue #282, opened 2016-10-07
Allow bar_format without total


pjvandehaar:
Thanks for `tqdm`!  It's a wonderful tool.

In order to use `bar_format`, `total` must be used.  I want to be able to do this without `total=int(1e100)`:

```
import tqdm, time
slow_iter = (time.sleep(1e-4) or i for i in range(int(1e5)))
with tqdm.tqdm(slow_iter, total=int(1e100), bar_format='Found {n:7} files') as prog_iter:
     for i in prog_iter:
          if i % int(1e4) == 0:
               prog_iter.write('Woohoo,{:6} files!'.format(i))
```

lrq3000:
Yes this will be implemented by #223 (I know it's a bit obscure, sorry about that, there are a lot of things in this PR). But thank you for your PR, we will merge it with #223.

PR #284, opened 2016-10-07
Allow bar_format without total or len(iterable)

pjvandehaar:
(This builds on https://github.com/tqdm/tqdm/pull/283 to allow me to test it locally.)

This PR is a solution to my issue here: https://github.com/tqdm/tqdm/issues/282 .


lrq3000:
Thank you @pjvandehaar, this will be merged with #223 which already implements bar_total when there is no total, but your PR has: an extended docstring (should also be added to the readme) + skip if coverall fails.

pjvandehaar:
Alright.  Feel free to just copy that docstring and reject this PR.

PR #326, opened 2016-12-23
sys.setcheckinterval (deprecated) changed to sys.setswitchinterval

deeenes:
`sys.setcheckinterval` (deprecated) changed to `sys.setswitchinterval`


lrq3000:
Thanks for the PR, note for us: we should enhance this PR to use the deprecated function for old Python versions and the new function for Python 3.

PR #330, opened 2016-12-31
Fix security issue in __version__ extra calculation (fix #328)

### LINKED ISSUE #328: tqdm._version: insecure use of git
When you import `tqdm`, the `tqdm._version` module executes the following command:
```
git log -n 1 --oneline
```
This was meant to check if the user is running a pre-release version of tqdm.
But most of the time there's no git repo at all, so this is just waste of time.
Worse, the current working directory might be a part of an unrelated git repository, possibly a malicious one.
At least with git 2.10 or later, it's possible to craft a repo in which `git log` executes arbitrary code:
```console
$ tail -n4 /tmp/.git/config
[log]
        showSignature = true
[gpg]
        program = /tmp/moogpg

$ tail -n4 /tmp/moogpg
#!/bin/sh
exec > /dev/tty 2>&1
cowsay pwned
sleep 9999

$ cd /tmp

$ pydoc tqdm
 _______
< pwned >
 -------
        \   ^__^
         \  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

lrq3000:
and add branch name in ```__version__``` if available.

This should fix the security issue reported in #328 (thanks to @jwilk!) alias CVE-2016-10075.

TODO (but can be done later, we should merge ASAP a fix for this issue):
- [ ] restore 100% coverage (need to virtualize .git folder and create 3 scenarios: no tqdm in config, head detached so commit hash directly in HEAD file, head in a branch so commit hash is in a ref file). Maybe use something like `with unittest.mock.patch.dict(os.environ, PATH='/nonexistent')` (courtesy of @jwilk)?

### DIFF
--- a/tqdm/_version.py
+++ b/tqdm/_version.py
@@ -11,35 +11,41 @@ __version__ = '.'.join(map(str, version_info))
 
 
 # auto -extra based on commit hash (if not tagged as release)
-res = None
 scriptdir = os.path.dirname(__file__)
 gitdir = os.path.abspath(scriptdir + '/../.git')
 if os.path.isdir(gitdir):
-    # Open the HEAD file
-    with open(os.path.join(gitdir,'HEAD'), 'r') as fh_head:
-        res = fh_head.readline().strip()
-    # If we are in a branch, HEAD will point to a file containing the latest commit
-    if 'ref:' in res:
-        # Get reference file path
-        ref_file = res[5:]
-        # Get branch name
-        branch_name = ref_file.split('/')[-1]
-        # Sanitize path of ref file
-        # get full path to ref file
-        ref_file_path = os.path.abspath(os.path.join(gitdir, ref_file))
-        # check that we are in git folder (by stripping the git folder from the ref file path)
-        if os.path.relpath(ref_file_path, gitdir).replace('\\', '/') != ref_file:
-            # Trying to get out of git folder, not good!
-            res = None
-        else:
-            # Else the file file is inside git, all good
-            # Open the ref file
-            with open(ref_file_path, 'r') as fh_branch:
-                commit_hash = fh_branch.readline().strip()
-                res = commit_hash[:8] + ' (branch: ' + branch_name + ')'
-    # Else we are in detached HEAD mode, we directly have a commit hash
-    else:
-        res = res[:8]
+    extra = None
+    # Open config file to check if we are in tqdm project
+    with open(os.path.join(gitdir, 'config'), 'r') as fh_config:
+        if 'tqdm' in fh_config.read():
+            # Open the HEAD file
+            with open(os.path.join(gitdir, 'HEAD'), 'r') as fh_head:
+                extra = fh_head.readline().strip()
+            # If we are in a branch, HEAD will point to a file containing the latest commit
+            if 'ref:' in extra:
+                # Get reference file path
+                ref_file = extra[5:]
+                # Get branch name
+                branch_name = ref_file.split('/')[-1]
+
+                # Sanitize path of ref file
+                # get full path to ref file
+                ref_file_path = os.path.abspath(os.path.join(gitdir, ref_file))
+                # check that we are in git folder (by stripping the git folder from the ref file path)
+                if os.path.relpath(ref_file_path, gitdir).replace('\\', '/') != ref_file:
+                    # Trying to get out of git folder, not good!
+                    extra = None
+                else:
+                    # Else the file file is inside git, all good
+                    # Open the ref file
+                    with open(ref_file_path, 'r') as fh_branch:
+                        commit_hash = fh_branch.readline().strip()
+                        extra = commit_hash[:8] + ' (branch: ' + branch_name + ')'
+
+            # Else we are in detached HEAD mode, we directly have a commit hash
+            else:
+                extra = extra[:8]  # limit to the first 8 symbols of the hash
+
     # Append to version string
-    if res is not None:
-        __version__ += '-' + res
+    if extra is not None:
+        __version__ += '-' + extra
--- a/tqdm/_version.py
+++ b/tqdm/_version.py
@@ -1,5 +1,6 @@
 # Definition of the version number
 import os
+from io import open as io_open
 
 __all__ = ["__version__"]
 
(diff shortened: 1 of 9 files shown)


tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
...
+    # If we are in a branch, HEAD will point to a file containing the latest commit
+    if 'ref:' in res:
+        # Get reference file path
+        ref_file = res[5:]
+        # Get branch name
+        branch_name = ref_file.split('/')[-1]
+        # Sanitize path of ref file
+        # get full path to ref file
+        ref_file_path = os.path.abspath(os.path.join(gitdir, ref_file))
+        # check that we are in git folder (by stripping the git folder from the ref file path)
```
CrazyPython: inconsistent capitalization of comments (PEP 8 recommends lowercase)

tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
...
-
-    cur_hash = commit_hash()
-    if cur_hash is not None:
-        last_release = commit_hash('v' + __version__)
-
-        if (last_release is None) or (cur_hash not in last_release):
-            __version__ += '-' + cur_hash
+res = None
+scriptdir = os.path.dirname(__file__)
+gitdir = os.path.abspath(scriptdir+'/../.git')
```
CrazyPython: no spaces around operator

tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
...
+            # Else the file file is inside git, all good
+            # Open the ref file
+            with open(ref_file_path, 'r') as fh_branch:
+                commit_hash = fh_branch.readline().strip()
+                res = commit_hash[:8] + ' (branch: ' + branch_name + ')'
+    # Else we are in detached HEAD mode, we directly have a commit hash
+    else:
+        res = res[:8]
+    # Append to version string
+    if res:
```
CrazyPython: ` is not None` can be more explicit

tqdm/_version.py
lrq3000: It's because the capitalized one pertains to the whole code block, whereas the lowercase pertains to only the next line (because there's not enough space to put it at the right of the code). I don't know how to make it more explicit or prettier...

lrq3000:
Thanks @CrazyPython, I fixed what you spotted :)

tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
 
 
 # auto -extra based on commit hash (if not tagged as release)
```
CrazyPython: PEP 8 dictates one less newline here

tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
...
-        last_release = commit_hash('v' + __version__)
-
-        if (last_release is None) or (cur_hash not in last_release):
-            __version__ += '-' + cur_hash
+res = None
+scriptdir = os.path.dirname(__file__)
+gitdir = os.path.abspath(scriptdir + '/../.git')
+if os.path.isdir(gitdir):
+    # Open the HEAD file
+    with open(os.path.join(gitdir,'HEAD'), 'r') as fh_head:
```
CrazyPython: Missing space after comma

tqdm/tests/tests_version.py
```diff
@@ -4,7 +4,8 @@
 def test_version():
     """ Test version string """
     from tqdm import __version__
-    Mmpe = re.split('[.-]', __version__)
```
CrazyPython: Not a very good variable name - CamelCase and undescriptive

tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
...
-        except:
-            return None
-
-    cur_hash = commit_hash()
-    if cur_hash is not None:
-        last_release = commit_hash('v' + __version__)
-
-        if (last_release is None) or (cur_hash not in last_release):
-            __version__ += '-' + cur_hash
+res = None
```
CrazyPython: `res` is defined here...

tqdm/_version.py
CrazyPython: ...but only used in this if statement. Either move it in (lazy evaluation) or move the checking operation out (imperative).

jwilk:
(Disclaimer: I haven't tested this; I only had a cursory look at the code.)
I think this is fine from security point of view.
However, if the user kept their local site-packages in a git repo, and installed tqdm there, you would be inspecting this unrelated repository, not the tqdm one. Oh well...
IMHO the desire of having super-exact `__version__` doesn't justify complexity of this code. I'd just remove all the git-related stuff from `_version.py`. But it's not my call...
Typo: `file file`.

tqdm/_version.py
CrazyPython: `r` is not required, you decide if that's a good idea

tqdm/_version.py
CrazyPython: You might want to rename `res` to `result`. It's more clear.

tqdm/_version.py
```diff
@@ -15,18 +11,35 @@
...
+scriptdir = os.path.dirname(__file__)
+gitdir = os.path.abspath(scriptdir + '/../.git')
+if os.path.isdir(gitdir):
+    # Open the HEAD file
+    with open(os.path.join(gitdir,'HEAD'), 'r') as fh_head:
+        res = fh_head.readline().strip()
+    # If we are in a branch, HEAD will point to a file containing the latest commit
+    if 'ref:' in res:
```
CrazyPython: In both of these branches, `res` is set. Yet above, `res` is set to another name. Two variables would be clearer.

tqdm/tests/tests_version.py
```diff
@@ -4,7 +4,8 @@
...
+    Mmpe = __version__.split()[0]  # remove part after space (branch name)
+    Mmpe = re.split('[.-]', Mmpe)  # split by dot and dash
     assert 3 <= len(Mmpe) <= 4
```
CrazyPython: Is this not the same as `assert len(Mmpe) in (3, 4)`? People usually expect reversed syntax, e.g. `4 >= len(Mmpe) >= 3`.

lrq3000:
Thanks both for your reviews and insights. I agree, we should think about
keeping this code.
Note to self: check git repo name (tqdm) and fix CrazyPython remarks.

tqdm/_version.py
lrq3000: I don't understand your remark, `res` always aim to contain the extra = the commit hash (and if available the branch name). I changed the variable name to be more explicit. It's just that if we are in a branch, we don't get the commit hash from the HEAD file, so then it's necessary to do some more processing.

tqdm/_version.py
lrq3000: yeah but better be explicit :)

tqdm/_version.py
lrq3000: Renamed to `extra`.

jwilk:
<= is waaay more common than >= in chained comparisons:
https://codesearch.debian.net/search?q=%3C%3D%5Cs%2B%5Cw%2B%5Cs%2B%3C%3D+filetype%3Apython (5363 hits)
https://codesearch.debian.net/search?q=%3E%3D%5Cs%2B%5Cw%2B%5Cs%2B%3E%3D+filetype%3Apython (78 hits)

lrq3000:
Wow regex in debian repo search, I did not know about that, amazing!

2017-01-01 15:59 GMT+01:00 Jakub Wilk <notifications@github.com>:

CrazyPython:
If I have time later, I might review some of the existing tqdm codebase, if you want me to.

lrq3000:
@CrazyPython yes please feel free to do so, your feedbacks are appreciated
:) But it might take you a while!

2017-01-01 21:16 GMT+01:00 CrazyPython <notifications@github.com>:

tqdm/tests/tests_version.py
```diff
@@ -4,9 +4,10 @@
...
     from tqdm import __version__
-    Mmpe = re.split('[.-]', __version__)
-    assert 3 <= len(Mmpe) <= 4
+    version_parts = __version__.split()[0]  # remove part after space (branch name)
+    version_parts = re.split('[.-]', version_parts)  # split by dot and dash
+    assert 3 <= len(version_parts) <= 4
     try:
-        map(int, Mmpe[:3])
+        map(int, version_parts[:3])
     except:
```
CrazyPython: You want `except ValueError:`. This catches "errors" like `SystemExit` and other python internals.

tqdm/_version.py
```diff
@@ -15,18 +11,41 @@
...
-
-        if (last_release is None) or (cur_hash not in last_release):
-            __version__ += '-' + cur_hash
+scriptdir = os.path.dirname(__file__)
+gitdir = os.path.abspath(scriptdir + '/../.git')
+if os.path.isdir(gitdir):
+    extra = None
+    # Open config file to check if we are in tqdm project
+    with open(os.path.join(gitdir, 'config'), 'r') as fh_config:
+        if 'tqdm' in fh_config.read():
```
sandrotosi: here you're just checking if 'tqdm' is anywhere in .git/config which might not be strict enough

lrq3000: @sandrotosi Yep I thought the same but after reading a bit about git internals, I didn't find a better way to check if we were looking at tqdm repository. Any suggestion? Maybe looking specifically for a specific entry in config file?

sandrotosi: i looks like you have to go thru several steps: read .git/HEAD, lookup the remote for that ref in .git/config, check the remote url in .git/config if that ends with '/tqdm.git' (as you have to consider forks and other possible checkout methods).

i still think there must be an easier way to do it, and i'd look at out other "big" projects (numpy, scipy, matplotlib, and the like) do as they probably want to tackle the same problem as you do

casperdcl:
I agree that it's a bit much to go overboard with this, it was meant to be a small helper. I'm tempted to ignore coverage of the "extra" version bit.

Issue #357, opened 2017-03-06
Readme doesn't list all bar_format options


postelrich:
The readme doesn't list all bar_format options.  I was using `set_postfix` on a progress bar that uses a custom bar format.  Noticed the postfix wasn't showing up.  Digging into the code, I noticed that it is also another kwarg that has to be defined in the format string.

postelrich:
Also noticed that if you want to use a custom bar format with the tqdm_notebook the format string needs to start with `{n}/|/`.  Had to glean this from the code.

PR #363, opened 2017-03-31
fix navigation for windows command shell

MPagel:
In addition to addressing documentation issues with the postfix keyword (#362 #266 #270 #347 #357), this patch changes the vertical navigation mode in windows from writing `u'\x1b[A'` to the console to using the windows API (`ctypes.windll.kernel32.SetConsoleCursorPosition(...)`). This is because `ANSI.sys` and the attendant support for escape sequence interpretation is not available in 64-bit versions Windows prior to Windows 10, rendering `colorama` and this repo ineffective. `ANSI.sys` was available until WindowsXP x86. Windows 10 has added support for `ENABLE_VIRTUAL_TERMINAL_PROCESSING` via `SetConsoleMode`. 

Currently I implement the windows 10 console mode shift via via pywin32's `win32console`, but it may also be supported in `ctypes.windll.kernel32` (but would require minor refactoring of variables and comparisons for success). However, `ENABLE_VIRTUAL_TERMINAL_PROCESSING` (0x0004) is not a recognized constant in the ctypes package.

Although windows version <= XP may not need this patch and this patch may slow down rendering, this patch should still function. In intermediate windows versions (e.g. XP-64, Vista, 7, 8, 8.1, early builds of 10), `SetConsoleCursorPosition` appears to be required for proper progress bar placement.

Future: detect if `ANSI.sys` (or `ansicon`) loaded, because in that case, ANSI escape sequences should work (rather than requiring ConsoleCursorPosition calls). Also not sure about powershell's handling of ANSI sequences and if that affects any diagnostic environmental variables.


casperdcl:
Thanks for lloking into this. I think the CI build is failing because it's executing your windows-specific code even though it's a linux CI.

MPagel:
I'll re-check the OS logic when I have a chance to do so in a couple of 
days.

MPagel:
Re: test failures.  Getting: AssertionError: 

```
Expected:
[u'\n\n\rpos2 bar:   0%', u'\x1b[A\x1b[A\n\n\rpos2 bar:  50%', u'\x1b[A\x1b[A\n\n\r      ', u'\x1b[A\x1b[A']

Got:
[u'\rpos2 bar:   0%', u'\rpos2 bar:  50%', u'\r      ']

Raw:
[u'\rpos2 bar:   0%|          | 0/2 [00:00<?, ?it/s]\rpos2 bar:  50%|#####     | 1/2 [00:00<00:00, 167772.16it/s]\r                                                           ']

```
this may be a failure simply because my recoded module doesn't always rely on ANSI sequences. There's stuff going on other than pure write-to-device.  But if this is truly being built in a linux shell, I didn't think any of my manipulations (for windows) should bugger up the regular positioning.

In other words, I'm at a loss as to 
a) how do we do a unit test on what amounts to a screen grab? and 
b) why the existing unit tests fail under a linux env.

Any input/feedback from the tqdm team would be appreciated at this point.

casperdcl:
Thanks @MPagel. Have a major backlog of `tqdm` stuff we need to look at but hopefully we'll get round to this soon.

MPagel:
finally passes Travis :P

MPagel:
(if you copy this to a branch of the main repo, see if you can rename it from enable_vt_escapes to windows_navigation or windows_non_ansi or something of that ilk)

MPagel:
now colorama works on my computer (windows 8.1x64), so I'll close this PR.  I'll leave my repo up on my end in case I find out some circumstances in which colorama doesn't work (or if people prefer a non-colorama solution).  I haven't tested to verify that vanilla tqdm works on my end yet, but I'll open a new pull request based on this code if I find some circumstances in which tqdm+colorama doesn't work.

casperdcl:
great, thanks @MPagel

PR #435, opened 2017-09-16
[Feature] microstep

sasdf:
Estimate progress of single iteration using average rate.  
It can make progress bar more responsive when single iteration takes very long to execute.

Usage:
```python
from tqdm import *

tqdm.monitor_interval = 0.1
bar = tqdm(total=5, maxinterval=0.1, microstep=True)
```
### Changes
`TMonitor` will refresh the bars every `maxinterval` seconds, updating progress, elapsed, and remaining.   

`tqdm.__iter__` will save `avg_time` back to instance attribute, it is needed for estimating current progress in `__repr__`.  

Two attribute added to `class tqdm`
1. `last_refresh_time_t`
    To record when the bar had been refreshed.  (we can't change `last_print_time_t` since it is needed for calculate `avg_time`)
2. `step_start_t`
    For calculate time elapsed of current iteration.

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -101,16 +101,21 @@ class TMonitor(Thread):
             cur_t = self._time()
             # Check for each tqdm instance if one is waiting too long to print
             # NB: copy avoids size change during iteration RuntimeError
+            #  print(len(self.tqdm_cls._instances))
             for instance in self.tqdm_cls._instances.copy():
-                # Only if mininterval > 1 (else iterations are just slow)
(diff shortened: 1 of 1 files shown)


lrq3000:
Thanks for the PR, very interesting feature particularly when progress is slow (step > 1s), but there is a lot of additional new self variables calls, we should check if the performance does not decrease too dramatically (also I expect we can enhance this a bit, particularly L979 https://github.com/tqdm/tqdm/pull/435/files#diff-35aa74e5cefc63c038a9a0f55b5166cdR979)

PR #477, opened 2017-11-07
Fix overlap of nested progressbar

dyno:
The problem is exhibitable with the modified examples/parallel_bars.py,
the fix is simple check pos is not None rather otherwise will fall into
the trap of 0.


dyno:
the problem is mentioned in https://github.com/tqdm/tqdm/pull/329.

PR #524, opened 2018-03-17
fix the problem #299 in terms of using tqdm in pandas

chengs:
- [x] rebase on to master
- [x] Add new test cases about pandas
- [x] fix #299 - find the correct `total` when using `progress_apply` and other functions in pandas.

#299 partially achieved its goal. With forbidding `*args`, one can get correct axis now, and use it to get the `total` (as number of iterations in `progress_apply`)

However, pandas sometimes runs several more iterations to optimise itself. For example, in `dataframe.progress_apply(func)`.

- if `func` is a `numpy.ufunc` or is very slow, then it will be executed `total` times.
- but if it is very quick or in some other cases, it will be executed `total+1` or `total+shape[axis0]` times.

I think there is no need to adjust such particular pandas implementation, so in the wrapper function, I write
```python
def wrapper(*args, **kwargs):
                    # update tbar correctly
                    # it seems pandas apply calls func twice on the first column/row 
                    # to decide whether it can take a fast or slow code path.
                    # so stop when t.total==t.n
                    t.update(n=1 if t.total and t.n < t.total else 0)
                    return func(*args, **kwargs)
``` 
if `total` is achieved, `tbar` will not be updated.

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -567,11 +567,10 @@ class tqdm(object):
 Except func, normal arguments are intentionally \
 not supported by (DataFrame|Series|GroupBy).progress_apply. \
 Use keyword arguments instead.""", fp_write=getattr(t.fp, 'write', sys.stderr.write))
-                
                 # Define bar updating wrapper
                 def wrapper(*args, **kwargs):
                     # update tbar correctly
-                    # it seems pandas apply calls func twice on the first column/row 
+                    # it seems pandas apply calls func twice on the first column/row
                     # to decide whether it can take a fast or slow code path.
                     # so stop when t.total==t.n
                     t.update(n=1 if t.total and t.n < t.total else 0)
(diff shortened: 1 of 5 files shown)


chengs:
I add new tests, so now, pandas.Series, pandas.Dataframe, and groupby cases are all covered. Not cover pandas.Panel because it is going to be removed. http://pandas.pydata.org/pandas-docs/version/0.20/whatsnew.html#whatsnew-0200-api-breaking-deprecate-panel

tqdm/_tqdm.py
```diff
@@ -539,21 +539,20 @@ def inner(df, func, *args, **kwargs):
                     Data (may be grouped).
                 func  : function
                     To be applied on the (grouped) data.
-                *args, *kwargs  : optional
+                **kwargs  : optional
                     Transmitted to `df.apply()`.
                 """
+
                 # Precompute total iterations
                 total = getattr(df, 'ngroups', None)
```
chengs: this `ngroups` is an inner attribute of pandas. It may be changed in future, but seems not happen in near future or never. So let's keep it so far. 

For perfection, I will suggest to use one of pandas public apis to get group size in future. So far, it is fine.

PR #541, opened 2018-04-16
Update document about changes in v4.23.0

chengs:
- now tqdm_notebook support ncols and dynamic 
- pandas (rolling|expanding).progress_apply


chengs:
@casperdcl @lrq3000 Also, I would like to apply for joining tqdm developers (https://github.com/orgs/tqdm/people). From recent applications, I think I showed that I am familiar with tqdm code and have the ability to contribute to it.

chengs:
@casperdcl look at this, this Travis-CI integration seems to have a problem. There is no way that this PL cannot pass the test.

README.rst
```diff
@@ -665,6 +693,11 @@ light blue: no ETA); as demonstrated below.
 |Screenshot-Jupyter2|
 |Screenshot-Jupyter3|
 
+Since v4.23.0, tqdm_notebook now can respect ``ncols`` and ``dynamic_ncols`` if 
+their values are given. Here is an example:
+
+|Screenshot-Jupyter4|
```
casperdcl: seems a bit much to add an extra screen shot just for this. would be nicer to include `ncols` in one of the examples above

chengs: Eh, the problem is I don't have gif generator... Or how about we just mention it and do not give an example?

chengs:
@casperdcl 
How about replacing the first GIF example by this new one?
tqdm_new

chengs:
@casperdcl if possible, could you check this PR?

chengs:
@casperdcl have you checked it?

PR #573, opened 2018-06-29
Fixed missing attr pos when used in multi-threaded enviornment

darindf:
Fixes attribute error on pos during decr_instances, uses similar approach as in #546


darindf:
Can this be merged, as several releases have been done with this being included

PR #590, opened 2018-08-09
[WIP] Handle outputs expecting bytes

altendky:
tqdm/tqdm#589

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -810,6 +810,12 @@ class tqdm(Comparable):
         if smoothing is None:
             smoothing = 0
 
+        try:
+            file.write(u'')
+        except TypeError:
+            import io
+            file = io.TextIOWrapper(file)
+
         # Store the arguments
         self.iterable = iterable
         self.desc = desc or ''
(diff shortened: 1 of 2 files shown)


altendky:
I think this actually fixed my issue but I've got some others now so I'll try to get that tidied up before I say this is really ready.  Regardless, at least you can look it over and see if it's going in a sensible direction or not.

Thanks for your time.

tqdm/_tqdm.py
```diff
@@ -810,6 +810,12 @@ def __init__(self, iterable=None, desc=None, total=None, leave=True,
         if smoothing is None:
             smoothing = 0
 
+        try:
+            file.write(u'')
```
Julian: This isn't a great way to check this (though this does look like a bug in tqdm to me).

E.g., the real `sys.stdout` on py2 will happily accept `u""`, but blow up on any actual non-ASCII unicode.

Seems to me that this is another instance of code that should be taking native strings, or else if it insists on writing unicode, yeah, it can't be writing unicode to sys.stdout/sys.stderr on py2.

PR #599, opened 2018-08-29
Fix tqdm_notebook description being truncated for long descriptions. #582

AbysmalBiscuit:
Fix tqdm_notebook description being truncated for long descriptions.
Related to issue #582

### DIFF
--- a/tqdm/_tqdm_notebook.py
+++ b/tqdm/_tqdm_notebook.py
@@ -108,6 +108,7 @@ class tqdm_notebook(tqdm):
 
         if desc:
             pbar.description = desc
+            pbar.style = {'description_width': 'initial'}
         # Prepare status text
         ptext = HTML()
         # Only way to place text to the right of the bar is to use a container
@@ -170,6 +171,7 @@ class tqdm_notebook(tqdm):
             # Update description
             if desc:
                 pbar.description = desc
+                pbar.style = {'description_width': 'initial'}
 
         return print_status
 


chengs:
Looks good to me.
One suggestion, this ``description_width`` seems to be added recently (in ipywidgets 7.1)
So it's better to verify first that this commit does not cause any error in previous ipywidgets versions.
If so, maybe to add a version detection line.

https://ipywidgets.readthedocs.io/en/stable/changelog.html

AbysmalBiscuit:
@chengs Thanks for the suggestions!

I tested to see if the impact of ipywidgets' version. The earliest release of ipywidgets that has `IntProgress` is 7.0.0a7,  trying to adjust the description width caused no issues for me (it just didn't do anything).
From version 7.0.0a10 adjusting the description width works.

The following are a series of srceenshots from the various relevant versions:
7.0.0a6 (no `IntProgress` yet):
screenshot from 2018-08-31 15-51-03

7.0.0a7 (`IntProgress` available but description width cannot be adjusted):
screenshot from 2018-08-31 15-51-23

7.0.0a9 (last release before description can be adjusted):
screenshot from 2018-08-31 15-51-54

7.0.0a10 (description can be adjusted):
screenshot from 2018-08-31 15-52-26

I don't know why with version 7.0.0a6 a `NameError`  wasn't thrown in `status_printer`.

Based on this, should I add a version check that will throw an `ImportError` for anything older than 7.0.0a7 and issue a warning when someone uses a version up to 7.0.0a10 (the warning being for: 7.0.0a7 <= version <= 7.0.0a9); or should I just throw an `ImportError` that tells people to install at least version 7.0.0a10?

chengs:
WOW, nice research work!
Regarding your question, IMHO, no need to remind user (throw an error or warning) on this issue. 
So it's better to check if ipywidget.version > 7.1, and only add ``{'description_width': 'initial'}`` if so.

AbysmalBiscuit:
Hey, so I had added a version check, and I realized a problem with it. Checking the version of ipywidgets requires that it be installed. 

Hence if `IntProgress` was imported from `IPython.html.widgets` (cases where IPython is old), then the check would throw a `NameError`.

I had tried assigning a description update function, based on the version of ipywidgets, and if it was installed, but that didn't work unfortunately.

Also, setting the style would only throw an error if a widget has no style attribute, otherwise it would just add the css rule to the page. The css rule by itself wouldn't do anything. I had tried to setup an environment using Anaconda to test older versions of IPython, but I couldn't get the jupyter notebook to run correctly.

chengs:
@AbysmalBiscuit I think, now it is okay.
Can you verify if it works (especially in python 2)?

AbysmalBiscuit:
@chengs sorry for the delay.

I have checked it using a Python2 kernel in Jupyter and it works fine.

casperdcl:
fixes #582

PR #600, opened 2018-09-01
Fix a typo in error message

lc0:
Was having some issues in google colab, there I found this error message with a typo


chengs:
eh, this is already https://github.com/tqdm/tqdm/pull/597

lc0:
ah, sorry for the duplicate! 

That moment, when having up to date master might help to avoid such duplicates :)

PR #615, opened 2018-09-17
Add 'fixed' option to fix the position of progress bars

mingrammer:
I've added `fixed` option to fix the position of progress bars. I think it is useful to make the position of multiple bars fixed while using multiprocessing.

With `fixed` option, you can fix the position of progress bars and make the bars keeping on its position even though a process was completed.

### Real case

I wanted to run 8 jobs with 3 processes with `multiprocessing`. As you know, an order of job running is unexpectable, so it may be challenging to keep the progress bars for each job on its fixed positions.

1. `leave=True` (default) and using `position`

`with tqdm(total=srcr.dbsize(), ascii=True, unit='keys', unit_scale=True, position=pos) as pbar:`

screenshot 2018-09-17 20 05 24

Yes it is tracing option, so, unexpected process switching are all traced. So I can't fix progress bars on a single line for a process.

2. `leave=False` and using `position`

`with tqdm(total=srcr.dbsize(), leave=False, ascii=True, unit='keys', unit_scale=True, position=pos) as pbar:`

screenshot 2018-09-17 20 05 45

With `leave=False`, each progress bar will be updated on its (fixed) position. But, when a progress is completed, bars will be disappeared. So, I could not see completed output.

3. `fixed=True` and using `position`

`with tqdm(total=srcr.dbsize(), fixed=True, ascii=True, unit='keys', unit_scale=True, position=pos) as pbar:`

screenshot 2018-09-17 20 42 40

With `fixed=True`, I could see completed outputs which were not disappeared.`fixed` option is very simillar to `leave=Fasle` except that it does not "clear" the completed progress bar (at `__exit()__`).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -356,6 +356,10 @@ Parameters
     Calls ``set_postfix(**postfix)`` if possible (dict).
 * unit_divisor  : float, optional  
     [default: 1000], ignored unless `unit_scale` is True.
+* fixed  : bool, optional
+    If [default: False], fix the progress bars on its position regardless of leave or termination.
+    If set it True, leave will be False automatically. (because it is meaningless when leave is True)
+    Useful to make the position of multiple bars fixed while using multiprocessing.
 
 Extra CLI Options
 ~~~~~~~~~~~~~~~~~
@@ -397,7 +401,7 @@ Returns
               [default: 1].
           """
 
-      def close(self):
+      def close(self. on_exit=False):
           """
           Cleanup and (if leave=False) close the progressbar.
           """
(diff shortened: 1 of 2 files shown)


chengs:
I help you fix the error in travis-ci. If you find the perf test reports an error, try to rerun it in travis-ci. Sometimes it happens.

As to the decision of whether use this PR, let's wait @casperdcl to make it.

mingrammer:
@chengs There seems no error in travis-ci

mingrammer:
@chengs Ah, that's good! Thank you :)

mingrammer:
Any progress?

PR #651, opened 2018-12-07
[Feature] Allow `total=float("inf")` 

RunOrVeith:
Treats iterables that are explicitly stated to be infinite the same as unknown length, instead of crashing with a TypeError or OverflowError.

Sample use case:

```
def infinite():
    i = 0
    while True:
        yield i

progress_bar = tqdm(infinite(), total=float("inf"))
```

In theory, this would also work with a class whose length is infinite.
However, this is currently not supported by python itself. (`__len__` must return an integer)

### DIFF
--- a/tqdm/_tqdm.py
+++ b/tqdm/_tqdm.py
@@ -702,8 +702,8 @@ class tqdm(Comparable):
             Prefix for the progressbar.
         total  : int, optional
             The number of expected iterations. If unspecified,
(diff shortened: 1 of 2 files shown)


chengs:
Does this also work for ``numpy.inf``?

RunOrVeith:
Yes, and also with math.inf, as they're all interchangeable (see [this so question](https://stackoverflow.com/questions/42315541/difference-between-np-inf-and-floatinf) for details)

PR #736, opened 2019-05-09
Provide itertools wrappers that know about their length.

anntzer:
A proof of concept for https://github.com/tqdm/tqdm/issues/225#issuecomment-490171284 (see discussion there); needs to be extended to more itertools iterators, needs docs.


anntzer:
Abandoned, but anyone interested should feel free to pick up the patch.

Issue #1000, opened 2020-07-06
position argument implementation


haji-ali:
- [ ] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [x] visual output bug
    + [ ] documentation request (i.e. "X is missing from the documentation." If instead I want to ask "how to use X?" I understand [StackOverflow#tqdm] is more appropriate)
    + [ ] new feature request
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [x] I have mentioned version numbers, operating system and
  environment, where applicable:
  ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)

  4.47.0 3.6.9 (default, Apr 18 2020, 01:56:04) 
  [GCC 8.4.0] linux
  ```

I have been having trouble with specifying the position argument for tqdm for a block of 8 bars. Consider the following code:

```python
from time import sleep
from tqdm import trange, tqdm

L = list(range(9))

def progresser(n):
    interval = 0.001 / (n + 2)
    total = 1000
    text = "#{}, est. {:<04.2}s".format(n, interval * total)
    for _ in trange(total, desc=text, position=n):
        sleep(interval)

if __name__ == '__main__':
    list(map(progresser, L))
    input(">")
```

This is the output:
![tqdm-output](https://user-images.githubusercontent.com/17233936/86609541-f6d5c280-bfa3-11ea-913a-c35a04172a14.gif)

As you can see the progress bars are not outputted correctly. I believe the issue is here https://github.com/tqdm/tqdm/blob/15c5c512fffca8fd98080cec322e8b6c197c0377/tqdm/std.py#L1293-L1294 

The problem is that when tqdm closes, the output is always in position 0 regardless of the value of `pos`. Another issue is that a new line `\n` is outputted. This is problematic because it means that the position for the next `tqdm` is no longer correct since the cursor is no longer at the beginning of the block.

Clearly this example is simplistic since the position argument is unnecessary. However it illustrates an unavoidable problem when using threading or multiprocessing.

One way to fix this would be to change `std.py:1293` to `self.display(pos=pos)` and to
have a tqdm command to update the output at the end of a tqdm block so that an appropriate number of new lines is outputted.

jpfeuffer:
I think the same or a similar problem exists for years already: https://github.com/tqdm/tqdm/issues/285
[...]

Issue #889, opened 2020-02-06
Multiprocessing example in docs doesn't work.


apengwin:
- [ ] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [ x] visual output bug
    + [ ] documentation request (i.e. "X is missing from the documentation." If instead I want to ask "how to use X?" I understand [StackOverflow#tqdm] is more appropriate)
    + [ ] new feature request
- [ x] I have visited the [source website], and in particular
  read the [known issues]
- [ x] I have searched through the [issue tracker] for duplicates
- [ x] I have mentioned version numbers, operating system and
  environment, where applicable:
  ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)
  ```

I want to use `tqdm` to show multiple concurrent progress bars similar to how `docker pull` shows the progress of parallel downloads concurrently. However, the simple multiprocessing example in the docs is buggy. Specifically, the `position` argument is not honored. The bars are first printed in correct order, and then re-printed in reverse order.

Here's the code i'm running
```
$ cat foo.py
from time import sleep
from tqdm import trange, tqdm
from multiprocessing import Pool, freeze_support

L = list(range(9))

def progresser(n):
    interval = 0.001 / (n + 2)
    total = 5000
    text = "#{}, est. {:<04.2}s".format(n, interval * total)
    for _ in trange(total, desc=text, position=n):
        sleep(interval)

if __name__ == '__main__':
    freeze_support()  # for Windows support
    p = Pool(initializer=tqdm.set_lock, initargs=(tqdm.get_lock(),))
    p.map(progresser, L)
$
```
 
Here are some screenshots to demonstrate.
<img width="1920" alt="Screen Shot 2020-02-05 at 5 42 09 PM (2)" src="https://user-images.githubusercontent.com/7637700/73898808-5d14b880-483f-11ea-87f4-3b0c22422081.png">
<img width="1920" alt="Screen Shot 2020-02-05 at 5 42 10 PM (2)" src="https://user-images.githubusercontent.com/7637700/73898819-6736b700-483f-11ea-84d4-b9a95246cc2d.png">
<img width="1920" alt="Screen Shot 2020-02-05 at 5 42 11 PM (2)" src="https://user-images.githubusercontent.com/7637700/73898826-6e5dc500-483f-11ea-8dfe-15eeec5b553a.png">

KingDarBoja:
Try this code instead:

```python
from time import sleep
[...]

PR #1001, opened 2020-07-09
Allow negative position arguments to fix parallel output issue

haji-ali:
    + [ ] exception-raising fix
    + [x] visual output fix
    + [x] new feature

  read the [known issues]
  ```python
  import tqdm, sys

  4.47.0 3.6.9 (default, Apr 18 2020, 01:56:04) 
  [GCC 8.4.0] linux
  ```

This commit allows the position argument in `tqdm` to be negative.  The reason is to fix the overlapping bars issues #285, #1000, #889 and https://github.com/tqdm/tqdm/pull/329#issuecomment-269557028

For example, with these changes the following example (using position `with_negative_pos = True`):
```python
#!/usr/bin/env python3
from time import sleep
from tqdm import trange, tqdm
from multiprocessing import Pool, freeze_support, Lock

L = list(range(9))

with_negative_pos = True

def progresser(n):
    interval = 0.001 / (n + 2)
    total = 1000
    pos = n
    if with_negative_pos:
        pos += -len(L)
    for _ in trange(total, desc=text, position=pos):
        sleep(interval)

if __name__ == '__main__':
    block = "\n"*len(L)*int(with_negative_pos)

    print("----------- Parallel Example: ")
    print(block, end='')
    p = Pool(initializer=tqdm.set_lock, initargs=(Lock(),))
    p.map(progresser, L)


    print("----------- Sequential Example: ")
    print(block, end='')
    list(map(progresser, L))

    input(">")
```
has the following output

new-negative

The idea is to ask the use to reserve the needed space (in lines) and then `tqmd` simply outputs the progress bar in one of the reserved lines.

Note that the output of the previous example (setting `with_negative_pos = False`) and without the changes proposed in this commit has the output:

old-positive

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -533,7 +533,7 @@ class tqdm(Comparable):
     @classmethod
     def _get_free_pos(cls, instance=None):
         """Skips specified instance."""
-        positions = set(abs(inst.pos) for inst in cls._instances
+        positions = set(inst.pos for inst in cls._instances
                         if inst is not instance and hasattr(inst, "pos"))
         return min(set(range(len(positions) + 1)).difference(positions))
 
@@ -565,7 +565,7 @@ class tqdm(Comparable):
                 if instances:
                     inst = min(instances, key=lambda i: i.pos)
                     inst.clear(nolock=True)
-                    inst.pos = abs(instance.pos)
+                    inst.pos = instance.pos
             # Kill monitor if no instances are left
             if not cls._instances and cls.monitor:
                 try:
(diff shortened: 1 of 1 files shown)


tqdm/std.py
```diff
@@ -1290,8 +1290,9 @@ def fp_write(s):
             if leave:
                 # stats for overall rate (no weighted average)
                 self.avg_time = None
-                self.display(pos=0)
-                fp_write('\n')
+                self.display(pos=pos)
```
haji-ali, line 1293: Except for this line, all changes in this commit do not change the output of previous the version of `tqdm` if `position` is positive. 
I am not sure why `pos=0` was used instead of `pos=pos` even when it is the position is positive. To completely preserve previous output, this line could be made
`self.display(pos=max(0,pos))`

PR #1005, opened 2020-07-15
Smoothing time

richardsheridan:
This PR provides a new setting, `smoothing_time` applicable for any use of tqdm. This new way to define smoothing is more general/universal; if you know how often you might glance at the progress bar, you know how many seconds to use for `smoothing_time`.

I always found the smoothing parameter hard to use and eventually settled on a recipe where I would set `smoothing = 1/(tqdm rate last time i did it)/seconds`, based on a little dimensional analysis and comparing EWMA formulas for discrete and time-based cases (e.g. Linux system load averages). This always seemed like a shame to do because obviously the numbers needed must be inside tqdm already. Lo and behold I finally go digging and the implementation is very elegant.

- Minimal performance penalty to old code (extra `elif self.smoothing_time:` statements)
- One extra division and one extra name assignment compared to `smoothing`
- Minimal new lines of code

When documenting the setting, I noticed a statement that smoothing is ignored in GUI mode. This doesn't seem to be the case for `tqdm.gui.tqdm` so I took the liberty of removing it.

I do apologize, but I couldn't get the test suite working on my local machine so I'll be relying on CI.

    + [ ] exception-raising fix
    + [ ] visual output fix
    + [x] new feature

  read the [known issues]
  ```python
  import tqdm, sys
  ```

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -846,9 +846,10 @@ class tqdm(Comparable):
     def __init__(self, iterable=None, desc=None, total=None, leave=True, file=None,
                  ncols=None, mininterval=0.1, maxinterval=10.0, miniters=None,
                  ascii=None, disable=False, unit='it', unit_scale=False,
-                 dynamic_ncols=False, smoothing=0.3, bar_format=None, initial=0,
-                 position=None, postfix=None, unit_divisor=1000, write_bytes=None,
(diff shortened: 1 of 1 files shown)


richardsheridan:
@casperdcl updated to work with new EMA class

PR #1007, opened 2020-07-15
Accidental strongref

richardsheridan:
This PR fixes a bug in `tqdm_tk` #1006 that prevents using the monitor thread. The thread was holding on to references to Tkinter objects longer than they lived in the main thread, and the `__del__` methods of said objects cannot be called from threads that don't host the `_tkinter.tkapp` instance. This falls afoul of a [known bug](https://bugs.python.org/issue39093) in CPython that I may fix later, and causes the whole interpreter to crash (even at the REPL), but for now this is a correctness issue of tqdm in its own right.

Apologies, but I haven't been able to get the test suite working on my machine so I hope CI will check it out.

    + [X] exception-raising fix
    + [ ] visual output fix
    + [ ] new feature

  read the [known issues]
  ```python
  import tqdm, sys
  ```

### DIFF
--- a/tqdm/_monitor.py
+++ b/tqdm/_monitor.py
@@ -90,12 +90,14 @@ class TMonitor(Thread):
                         instance.miniters = 1
                         # Refresh now! (works only for manual tqdm)
                         instance.refresh(nolock=True)
-                # Remove accidental long-lived strong reference
-                del instance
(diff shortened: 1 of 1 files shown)


richardsheridan:
This was a crazy bug to find... Everything works fine the first time you call, but then it crashes hard 10 seconds into the next call ...

PR #1054, opened 2020-10-21
Multiple bars display fix

### LINKED ISSUE #1000: position argument implementation

kolayne:
# About
Fixes several issues with bars positioning when using multiple bars. Resolves #1000 and probably some other similar issues.

# Other solutions
I know there already exists a PR fixing the problem (#1001), but I don't like it for a few reasons, including:
-  it simply doesn't work for me (however, I didn't do my best to make it. I'll probably try it later and tell about my results on its page)
-  it introduces negative positions, which I don't find intuitive, and (if I'm not mistaken) requires the library users to change their code to make it work correctly

# Changed and unchanged behaviour
Here there are samples I was running to compare the modified `tqdm` with the original one. All the samples were run on Python with `sys.platform == 'linux'`, `sys.version == '3.8.5 (default, Jul 28 2020, 12:59:40) \n[GCC 9.3.0]'`. The "before" screenshots are run on `tqdm` installed from `master`, the "after" screenshots are run on `tqdm` installed from the current branch (`bars_position_fix`).
<details>
<summary>Trivial</summary>

(Just to verify I haven't broken the very basics)
```Python3
from time import sleep

from tqdm import trange


for _ in trange(10):
    sleep(0.1)

for _ in trange(10, leave=False):
    sleep(0.1)
```
## After

</details>
<details>
<summary>Nested loops</summary>

```python3
from time import sleep

from tqdm import trange


for _ in trange(10):
    for _ in trange(10):
        sleep(0.1)
```
## After

</details>
<details>
<summary>Multithreaded with `position`</summary>

Please, pay attention to the order of the bars (numbers in the beginnings of the lines): in the version before the PR bars change their order. Sometimes (when using enough many bars or running them long enough) a similar sample can also produce empty lines or duplicate bars (if you want, I can try to dig into it and provide such a sample)
```python3
from time import sleep
from threading import Thread

from tqdm import trange


def progresser(n):
    interval = 0.001 / (n + 2)
    total = 1000
    for _ in trange(total, desc=text, position=n):
        sleep(interval)


if __name__ == '__main__':
    list(map(lambda x: Thread(target=progresser, args=(x,)).start(), range(9)))
```
## After

</details>
<details>
<summary>One-threaded with `position` and `leave=False`</summary>

```python3
from time import sleep

from tqdm import trange


def progresser(n):
    interval = 0.001 / (n + 2)
    total = 1000
    for _ in trange(total, desc=text, position=n, leave=False):
        sleep(interval)


if __name__ == '__main__':
    list(map(progresser, range(9)))
```
## After

</details>
<details>
<summary>One-threaded with `position`</summary>

This is the only case (I can think of) where you have to change code to make it produce a perfect output (if not changing it, the produced output is still better, than the output before the PR, as for me). However, I don't think this is a serious issue, because I think, this case is extremely special: there is no need to use bars with `position` if the application is one-threaded. In theory, there is one more situation where this case may come (when you use a multi-threaded application, but for some reason, a subset of the bars you want to use both starts and closes before other bars are created).

```python3
from time import sleep

from tqdm import trange


def progresser(n):
    interval = 0.001 / (n + 2)
    total = 1000
    for _ in trange(total, desc=text, position=n):
        sleep(interval)


if __name__ == '__main__':
    list(map(progresser, range(9)))
```
## After

Updated sample:
```python3
from time import sleep

from tqdm import trange


def progresser(n):
    interval = 0.001 / (n + 2)
    total = 1000
    for _ in trange(total, desc=text, position=n, terminate_lines=False):
        sleep(interval)


if __name__ == '__main__':
    list(map(progresser, range(9)))
    print('\n' * 9, end='')
```
## After edit

</details>

# Tests
I have very little experience with tests (both in python and generally). My PR has obviously changed `tqdm` outputs in some cases, so I think some tests will fail and I'll need help from maintainers (@casperdcl?) to understand, where I should change something back and where tests should be updated instead

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -1303,8 +1303,8 @@ class tqdm(Comparable):
             if leave:
                 # stats for overall rate (no weighted average)
                 self._ema_dt = lambda: None
-                self.display(pos=0)
-                fp_write('\n')
+                self.display()
+                fp_write('\r')
             else:
                 # clear previous display
                 if self.display(msg='', pos=pos) and not pos:
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -574,6 +574,19 @@ class tqdm(Comparable):
                          " (monitor_interval = 0) due to:\n" + str(e),
                          TqdmMonitorWarning, stacklevel=2)
                     cls.monitor_interval = 0
+
+            # Create simultaneous bars counters field
+            if not hasattr(cls, '_simultaneous_bars_counters'):
+                # To make multiple bars display correctly but also let
+                # user program safely print something after all the
+                # `tqdm` bars are closed, we only print terminate
+                # lines after the very last bar occupying the file
+                # closes.
+                # The following dictionary's keys are files, values
+                # are numbers of bars in the corresponding file we
+                # want to skip with `'\n'`s when the last bar closes
+                cls._simultaneous_bars_counters = {}
+
         return instance
 
     @classmethod
@@ -593,25 +606,26 @@ class tqdm(Comparable):
         order is not maintained but screen flicker/blank space is minimised.
         (tqdm<=4.44.1 moved ALL subsequent unfixed bars up.)
         """
-        with cls._lock:
-            try:
-                cls._instances.remove(instance)
-            except KeyError:
-                # if not instance.gui:  # pragma: no cover
-                #     raise
-                pass  # py2: maybe magically removed already
-            # else:
-            if not instance.gui:
-                last = (instance.nrows or 20) - 1
-                # find unfixed (`pos >= 0`) overflow (`pos >= nrows - 1`)
-                instances = list(filter(
-                    lambda i: hasattr(i, "pos") and last <= i.pos,
-                    cls._instances))
-                # set first found to current `pos`
-                if instances:
-                    inst = min(instances, key=lambda i: i.pos)
-                    inst.clear(nolock=True)
-                    inst.pos = abs(instance.pos)
+        # The lock is acquired by the caller of the method
+
+        try:
+            cls._instances.remove(instance)
+        except KeyError:
+            # if not instance.gui:  # pragma: no cover
+            #     raise
+            pass  # py2: maybe magically removed already
+        # else:
+        if not instance.gui:
+            last = (instance.nrows or 20) - 1
+            # find unfixed (`pos >= 0`) overflow (`pos >= nrows - 1`)
+            instances = list(filter(
+                lambda i: hasattr(i, "pos") and last <= i.pos,
+                cls._instances))
+            # set first found to current `pos`
+            if instances:
+                inst = min(instances, key=lambda i: i.pos)
+                inst.clear(nolock=True)
+                inst.pos = abs(instance.pos)
 
     @classmethod
     def write(cls, s, file=None, end="\n", nolock=False):
@@ -875,8 +889,8 @@ class tqdm(Comparable):
             If `None`, will leave only if `position` is `0`.
         file  : `io.TextIOWrapper` or `io.StringIO`, optional
             Specifies where to output the progress messages
-            (default: sys.stderr). Uses `file.write(str)` and `file.flush()`
-            methods.  For encoding, see `write_bytes`.
+            (default: sys.stderr). Uses `file.write(str)`, `file.flush()`
+            and `hash(file)`. For encoding, see `write_bytes`.
         ncols  : int, optional
             The width of the entire output message. If specified,
             dynamically resizes the progressbar to stay within this bound.
@@ -1107,6 +1121,12 @@ class tqdm(Comparable):
             else:  # mark fixed positions as negative
                 self.pos = -position
 
+        if leave and not disable:
+            with self._lock:
+                self._simultaneous_bars_counters[file] = \
+                    max(abs(self.pos) + 1,
+                        self._simultaneous_bars_counters.get(file, 0))
+
         if not gui:
             # Initialize the screen printer
             self.sp = self.status_printer(self.fp)
@@ -1274,9 +1294,7 @@ class tqdm(Comparable):
         # Prevent multiple closures
         self.disable = True
 
-        # decrement instance pos and remove from internal set
         pos = abs(self.pos)
-        self._decr_instances(self)
 
         if self.last_print_t < self.start_t + self.delay:
             # haven't ever displayed; nothing to clear
@@ -1300,6 +1318,9 @@ class tqdm(Comparable):
         leave = pos == 0 if self.leave is None else self.leave
 
         with self._lock:
+            # decrement instance pos and remove from internal set
+            self._decr_instances(self)
+
             if leave:
                 # stats for overall rate (no weighted average)
                 self._ema_dt = lambda: None
@@ -1307,8 +1328,24 @@ class tqdm(Comparable):
                 fp_write('\r')
             else:
                 # clear previous display
-                if self.display(msg='', pos=pos) and not pos:
+                if self.display(msg='', pos=pos):
                     fp_write('\r')
+
+            # No matter if current bar's traces should be left or not,
+            # if it's the last bar, it should print write terminators
+            if self.fp in self._simultaneous_bars_counters.keys():
+                for instance in self._instances:
+                    # If this is not the last bar, do nothing
+                    if instance.fp == self.fp:
+                        break
+                else:
+                    # Terminate lines to make further writes to the file
+                    # possible
+                    fp_write('\n' *
+                             self._simultaneous_bars_counters[self.fp])
+                    # Reset counter for current file (because the
+                    # previous bars are skipped already)
+                    del self._simultaneous_bars_counters[self.fp]
 
     def clear(self, nolock=False):
         """Clear current bar display."""
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -865,8 +865,8 @@ class tqdm(Comparable):
                  ascii=None, disable=False, unit='it', unit_scale=False,
                  dynamic_ncols=False, smoothing=0.3, bar_format=None, initial=0,
                  position=None, postfix=None, unit_divisor=1000, write_bytes=None,
-                 lock_args=None, nrows=None, colour=None, delay=0, gui=False,
-                 **kwargs):
+                 lock_args=None, nrows=None, colour=None, delay=0,
+                 terminate_lines=True, gui=False, **kwargs):
         """
         Parameters
         ----------
@@ -975,6 +975,17 @@ class tqdm(Comparable):
             Bar colour (e.g. 'green', '#00ff00').
         delay  : float, optional
             Don't display until [default: 0] seconds have elapsed.
+        terminate_lines  : bool, optional
+            Whether to print empty lines after the bars when all of them
+            are finished.
+
+            At the moment when the current bar gets finished: if there
+            are other bars left, the value is ignored; otherwise:
+            if [default: True], there'll be empty lines printed such
+            that the cursor is put below all the left bars, so that
+            you can print something below, if `False`, the cursor is
+            put back where it was before the group of bars has started,
+            so that you can overwrite them.
         gui  : bool, optional
             WARNING: internal parameter - do not use.
             Use tqdm.gui.tqdm(...) instead. If set, will attempt to use
@@ -1102,6 +1113,7 @@ class tqdm(Comparable):
         self.bar_format = bar_format
         self.postfix = None
         self.colour = colour
+        self.terminate_lines = terminate_lines
         self._time = time
         if postfix:
             try:
@@ -1339,13 +1351,14 @@ class tqdm(Comparable):
                     if instance.fp == self.fp:
                         break
                 else:
-                    # Terminate lines to make further writes to the file
-                    # possible
-                    fp_write('\n' *
-                             self._simultaneous_bars_counters[self.fp])
-                    # Reset counter for current file (because the
-                    # previous bars are skipped already)
-                    del self._simultaneous_bars_counters[self.fp]
+                    if self.terminate_lines:
+                        # Terminate lines to make further writes to the
+                        # file possible
+                        fp_write('\n' *
+                                 self._simultaneous_bars_counters[self.fp])
+                        # Reset counter for current file (because the
+                        # previous bars are skipped already)
+                        del self._simultaneous_bars_counters[self.fp]
 
     def clear(self, nolock=False):
         """Clear current bar display."""
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -602,12 +602,12 @@ class tqdm(Comparable):
         Remove from list and reposition another unfixed bar
         to fill the new gap.
 
+        This method must be called with `self._lock` acquired
+
         This means that by default (where all nested bars are unfixed),
         order is not maintained but screen flicker/blank space is minimised.
         (tqdm<=4.44.1 moved ALL subsequent unfixed bars up.)
         """
-        # The lock is acquired by the caller of the method
-
         try:
             cls._instances.remove(instance)
         except KeyError:
@@ -1135,6 +1135,11 @@ class tqdm(Comparable):
 
         if leave and not disable:
             with self._lock:
+                # The number of lines to print after this bar is added is:
+                # Either position of this bar + 1 (because of 0-indexing),
+                # in case the current bar's position is greater than
+                # position of all the other bars,
+                # Or the old value, in case this bar has a higher position
                 self._simultaneous_bars_counters[file] = \
                     max(abs(self.pos) + 1,
                         self._simultaneous_bars_counters.get(file, 0))
@@ -1344,7 +1349,7 @@ class tqdm(Comparable):
                     fp_write('\r')
 
             # No matter if current bar's traces should be left or not,
-            # if it's the last bar, it should print write terminators
+            # if it's the last bar, it should print line terminators
             if self.fp in self._simultaneous_bars_counters.keys():
                 for instance in self._instances:
                     # If this is not the last bar, do nothing
--- a/tqdm/utils.py
+++ b/tqdm/utils.py
@@ -171,6 +171,9 @@ class DisableOnWriteError(ObjectWrapper):
     def __eq__(self, other):
         return self._wrapped == getattr(other, '_wrapped', other)
 
+    def __hash__(self):
+        return hash(self._wrapped)
+
 
 class CallbackIOWrapper(ObjectWrapper):
     def __init__(self, callback, stream, method="read"):


tqdm/std.py
```diff
@@ -1329,16 +1355,36 @@ def fp_write(s):
...
                 # stats for overall rate (no weighted average)
                 self.avg_time = None
-                self.display(pos=0)
-                fp_write('\n')
+                self.display()
+                fp_write('\r')
             else:
                 # clear previous display
-                if self.display(msg='', pos=pos) and not pos:
+                if self.display(msg='', pos=pos):
```
kolayne, line 1368: I am not 100% sure about this change. I don't understand what is `and not pos` here for, but it affects the fourth example from the PR description (with `leave=False`): it leaves an unexpected newline symbol

tqdm/std.py
```diff
@@ -1307,9 +1335,7 @@ def close(self):
         # Prevent multiple closures
         self.disable = True
 
-        # decrement instance pos and remove from internal set
         pos = abs(self.pos)
-        self._decr_instances(self)
```
kolayne, line 1312: Have to combine the locking inside of `_decr_instances` and the locking located some lines below to avoid a race condition, which may result in 2 (or more) bars terminating lines, while only the very last one must

haji-ali:
This does fix the issue that I raised (when I use `terminate_lines`) and wrote the negative-position patch for. 
However, I don't understand why is it impossible for the case "One-threaded with `position`" to work correctly without modifications to user code? That seems like a standard case that `tqdm` should be able to handle.

Also, the nested loop example is still strange IMO. I would expect based on the documentation that all 10 bars should remain after execution (since `leave` is true) with another top bar for the outer loop.

kolayne:
@haji-ali, thank you for the constructive feedback!

> However, I don't understand why is it impossible for the case "One-threaded with `position`" to work correctly without modifications to user code? That seems like a standard case that `tqdm` should be able to handle.

I expect `tqdm` to move the cursor below all the bars after they close, to let me print something to the file. So, when the only bar is running, `tqdm` moves the cursor immediately after its completion. So, `pos` of the next bar is applied when the cursor is moved already, that's why empty lines appear.
But even if I'd change this behavior (i. e. make `terminate_lines=True` by default), the end-user will anyway have to either tell `tqdm` that all the bars are completed to make `tqdm` move the cursor or move it manually, to be able to print something later, so I don't think it's possible to avoid code changes in such case.

> Also, the nested loop example is still strange IMO. I would expect based on the documentation that all 10 bars should remain after execution (since `leave` is true) with another top bar for the outer loop.

I don't think, this is a feature because the following sample produces a similar output for the current version of `tqdm` in master (which surely is a bug):
```python3
for _ in trange(10, position=0):
    for _ in trange(10, position=1):
        pass
```
I think this following part from the docs is supposed to tell that the sample from the PR's description should be interpreted just like the above sample (however, that's not obvious at all, I'd like to get a confirmation or a denial from @casperdcl):
https://github.com/tqdm/tqdm/blob/8f9e03d1cf03db41ff9790091831d9583c7f229f/tqdm/std.py#L918-L920
But even if it's not a bug, it should be fixed in a separate PR, because this is an issue with the `_get_free_pos` method, used to figure out the position of a bar automatically:
https://github.com/tqdm/tqdm/blob/8f9e03d1cf03db41ff9790091831d9583c7f229f/tqdm/std.py#L1079

Knaifhogg:
I tried this PR with my [nested multi-thread long function](https://github.com/Knaifhogg/files/blob/main/nested_long_function_progress.py) that I've been having issues with, and it seems to resolve most of my problems.
Setup: 
- Windows 10 64-bit
- Python 3.8.236
- tqdm 4.51.0
- running in cmd

Before:
image
After:
image

Now I only have to figure out a way to count up the inner loop time spent before it is completed.

kolayne:
Hello, @casperdcl. I'm sorry for bothering you, but I wonder if there is any news on this?

casperdcl:
Ok finally reviewed this (!) Referring to the screenshots:

> Trivial

ok

> Nested loops

this is identical to setting `leave=False` for nested loops. That was the original nested implementation but it was changed to the current behaviour because more people wanted that.

> Multithreaded with `position`

Same as above - bars are "left" (i.e. printed like normal lines) as they are completed.

> One-threaded with `position` and `leave=False`

The whitespace causing a newline in the current version is odd - looks like a bug.

> One-threaded with `position`

I'm not too concerned by this as both versions look bad to me. If someone manually specifies position in a naïve way (both cases) then the output will look bad.

kolayne:
@casperdcl, thanks! :)
Here are my clarifications / questions


> this is identical to setting `leave=False` for nested loops. That was the original nested implementation but it was changed to the current behaviour because more people wanted that

Oops, looks like I misunderstood the `tqdm.__init__`'s docstring. I guess I thought the default value of `leave` is `None`. I agree that the new behaviour is a bug. But the behaviour in `master` at the moment is a bug too, isn't it? If what you're saying is you don't like my PR is replacing one bug with another one, then I'm going to work on it and fix it to look in the way you want it to (how do you want it to, btw? Should the bar of the outer loop be first and the other bars be printed below one after another? I strongly believe this should be documented somewhere in `tqdm.__init__` (in the section explaining the automatic position determination) or in `tqdm._get_free_pos`), but if you are saying the master's behaviour is not a bug (i.e. `tqdm` is expected to work this way), I think most users would disagree with you.


> Same as above - bars are "left" (i.e. printed like normal lines) as they are completed

???
Are you saying it is the expected behaviour for bars to change their positions during the execution? Hope, no, but then I don't understand what you meant. The set of bars printed before and after the PR/patch is the same, but my PR fixes the order. Please, add more details


> I'm not too concerned by this as both versions look bad to me. If someone manually specifies position in a naïve way (both cases) then the output will look bad.

Not sure if I understood you correctly. Which both of the three (Before, After, After edit) are you talking about?

casperdcl:
```python
from time import sleep
print(sleep(1), "task 3 completed")
print(sleep(1), "task 2 completed")
print(sleep(1), "task 1 completed")
```

will print statements in order or completion (3, 2, 1). The same is true of any bar which has `leave=True`.

```python
for i in tqdm([1], desc="task 1"):
  for j in tqdm([1], desc="task 2"):
    for k in tqdm([1], desc="task 3"):
      sleep(1)
```

The final order is (3, 2, 1).

The fact that while the bars are "running" the transient order is (1, 2, 3) is besides the point. If you would like to maintain this transient order even after the bars have closed (i.e. print final order (1, 2, 3)), this is identical to expecting `print(3); print(2); print(1)` to print in reverse order. While this is do-able, it's up to the end user to implement their own context manager/hack to get this highly non-standard behaviour.

kolayne:
> this is identical to expecting `print(3); print(2); print(1)` to print in reverse order. While this is do-able, it's up to the end user to implement their own context manager/hack to get this highly non-standard behaviour.

This is questionable. In my opinion, the standard behaviour for bars is to not be overwritten by other bars if `leave=True`. It sounds more standard if the bars' positions aren't messed up during the process, doesn't it?
And whether you agree with me or not, because there is a misinterpretation(?), this should be documented too!

haji-ali:
I personally find the behavior with the patch more intuitive. The nested loop behavior is especially confusing. The bars being overwritten and unfinished bars being left over makes it difficult to tell the progress if I am not following how the new bars are being created.

kolayne:
Looks like the discussion is stuck. Is there some well-established mechanism to find out the opinion of users on such questions? How do we figure out which of these do people _actually_ find more intuitive?

kolayne:
@casperdcl?..
> Looks like the discussion is stuck. Is there some well-established mechanism to find out the opinion of users on such questions? How do we figure out which of these do people _actually_ find more intuitive?

espdev:
I have checked this PR. If we do not set position argument and use a semaphore, the code works incorrectly.

The example code:
```python
from concurrent.futures import ThreadPoolExecutor, wait
from threading import Semaphore
import time
import random

from tqdm import tqdm

def worker(pos, sem):
    t = random.random() * 0.05
    with sem:
        # for _ in tqdm(range(100), desc=f'pos {pos}', position=pos):  # it works correctly
        for _ in tqdm(range(100), desc=f'pos {pos}'):  # it works incorrectly
            time.sleep(t)

def main():
    with ThreadPoolExecutor() as executor:
        sem = Semaphore(3)
        futures = []
        for pos in range(10):
            future = executor.submit(worker, pos, sem)
            futures.append(future)

        wait(futures)

if __name__ == '__main__':
    main()
```

The new bars overwrite the old bars if we do not set `position` argument:

tqdm_multi_5

BenjaminChoou:
[You may check this.](https://github.com/tqdm/tqdm/issues/1000#issuecomment-848837562)

kolayne:
Hello! I'm back here. Sincerely sorry for such a long delay, I start getting closer to removing the Busy mark from my profile.

> I have checked this PR. If we do not set position argument and use a semaphore, the code works incorrectly.

@espdev, thank you so much for pointing it out. I'm going to fix this (hopefully, soon).

In the model which I propose in this PR, this bug seems to be caused by the logic of `_get_free_pos`, which I didn't alter with the other changes. Before I can fix it, I'd like to find out (from @casperdcl, most likely): is that a bug or a feature that the `tqdm._get_free_pos` function does not do anything with the `file` argument of `tqdm.__init__`? Should it find the highest position _in the file_ where the new bar is created instead of what it does now (end-to-end position numbering across files)?

Method's source:
https://github.com/tqdm/tqdm/blob/140c94855bf98e7e0fda55b040a7a2c8ac76e786/tqdm/std.py#L579-L584

Example code:
```python
from time import sleep
from threading import Thread

from tqdm import tqdm


def run_me(f):
    for _ in tqdm(list(range(20)), file=f):
        sleep(0.1)


if __name__ == "__main__":
    f = open("/tmp/f1", "w")
    t1 = Thread(target=run_me, args=(f,))
    t2 = Thread(target=run_me, args=(None,))
    t1.start()
    t2.start()
    t1.join()
    t2.join()
```

Result of the code execution: https://asciinema.org/a/431831

kolayne:
@casperdcl, I have one more question regarding the expected bars' behavior. By the reaction you have left I can tell that you agree that the behavior found by espdev is incorrect. I assume that in that case it was expected that new bars appear under the completed ones. But I wonder about one more complicated case. Imagine there were three bars created: first with `leave=True`, second with `leave=False`, third with `leave=True`. At some point, the second bar closes, and after that, a new bar with `leave=True` is created. Should the third bar move to the place of the second one when the second gets closed (i.e. it collapses and other bars move up)? Should the third bar just leave an empty line so that the fourth will take its place when it's created? Or should the fourth bar take the fourth line (so that the second line will remain empty forever)?

BenjaminChoou:
@kolayne 

So far, I use Numba to help me achieve multi-progress bars display with multi-threads. But this cannot be implemented via the python multi-process library since communication between different processes is not straight.

Therefore, I wonder if it is possible for tqdm to maintain an internal pool for multi-progress bars in the master process while other processes can send a signal to update those bars. Then it would be easier for users to manage the multiple progress bars not only for multi-thread usage but also for complicated scenarios.

Please check my naive implementation.

```

from tqdm import tqdm
import time


class ProgressBar(object):
    def __init__(self, name, descs, totals):
        self.name = name
        self.bars = [tqdm(total=total, desc=desc, position=i)
                     for i, (desc, total) in enumerate(zip(descs, totals))]
        self.end_time = [0.0 for _ in range(len(totals))]

    def reset_start_time(self, bar_id):
        self.bars[bar_id].last_print_t = time.time()
        self.bars[bar_id].start_t = self.bars[bar_id].last_print_t

    def update(self, bar_id: int):
        assert 0 <= bar_id < len(self.bars)

        this_bar = self.bars[bar_id]
        cur_progress = this_bar.n
        total = this_bar.total

        if cur_progress < total:
            this_bar.update(1)
            if cur_progress == total - 1:
                this_bar.refresh()
            self.end_time[bar_id] = this_bar.last_print_t

    def close(self):
        for bar_id, bar in enumerate(self.bars):
            # reset the position of bar to 0 before close
            # so it won't be in wrong position
            bar.pos = 0
            # correct the time elapsed
            bar.start_t += (time.time() - self.end_time[bar_id])
            bar.close()


class ProgressBarManager(object):
    """
    A manager class for multiple progress bars.
    Potential usage is to display progress bars for multi-thread applications.
    """
    pool = None

    def __init__(self):
        raise RuntimeError('`ProgressBarPools` is not allowed to be instantiated')

    @staticmethod
    def create_pool(name, descs, totals):
        if ProgressBarManager.pool is None:
            ProgressBarManager.pool = ProgressBar(name, descs, totals)
        else:
            raise RuntimeError(
                'Progressbar pool should be empty before getting new pool:`%s` exists' % ProgressBarManager.pool.name)

    @staticmethod
    def reset_start_time(bar_id):
        assert isinstance(ProgressBarManager.pool, ProgressBar)
        ProgressBarManager.pool.reset_start_time(bar_id)

    @staticmethod
    def update(bar_id):
        assert isinstance(ProgressBarManager.pool, ProgressBar)
        ProgressBarManager.pool.update(bar_id)

    @staticmethod
    def get_total(bar_id):
        assert isinstance(ProgressBarManager.pool, ProgressBar)
        return ProgressBarManager.pool.bars[bar_id].total

    @staticmethod
    def close():
        assert isinstance(ProgressBarManager.pool, ProgressBar)
        ProgressBarManager.pool.close()
        ProgressBarManager.pool = None


create_pool = ProgressBarManager.create_pool
reset_start_time = ProgressBarManager.reset_start_time
update = ProgressBarManager.update
get_total = ProgressBarManager.get_total
close = ProgressBarManager.close

import numba
from numba import objmode
import time


@numba.jit(nopython=True, parallel=True)
def multi_thread_test():
    n_jobs = min(numba.get_num_threads(), 8)

    with objmode():
        # run code block by python interpreter
        descs = ['Batch # %d' % i for i in range(n_jobs)]
        totals = [30 * (i + 1) * (n_jobs - i) for i in range(n_jobs)]

        create_pool('Test', descs, totals)

    for batch_id in numba.prange(n_jobs):
        # run loops in parallel
        total = 30 * (batch_id + 1) * (n_jobs - batch_id)

        with objmode():
            reset_start_time(batch_id)

        for i in range(total):
            with objmode():
                time.sleep(0.01)
                update(batch_id)

    with objmode():
        close()


multi_thread_test()
```

BenjaminChoou:
> > Therefore, I wonder if it is possible for tqdm to maintain a unique internal pool for multi-progress bars in the master process while other processes can send a signal to update those bars.
> 
> I believe the solutions to the problem exist, but I consider them going out of the frame of this PR: here I only focus on problems with bars positioning caused by logical bugs and don't focus on processes' synchronization. You might want to discuss this in another (a new one?) issue/PR.
> 
> > Then it would be easier for users to manage the multiple progress bars not only for multi-thread usage but also for complicated scenarios.
> 
> To be honest, I wouldn't say the implementation you've provided is _easy_ to use: I see that you need to manually call the `update` method. And, in my opinion, it's one of `tqdm`'s key features that you don't need to modify code anyhow but just surrounding `for` loops with `tqdm()` to make it work.
> 
> > Please check my naive implementation.
> 
> I am not familiar with `numba`, so I can't tell exactly what the code does, although I don't see any place where any synchronization is performed, so the only difference from using usual `tqdm`s in multiple processes, which I see in your code, is explicitly specifying `position`s. Besides, if I was doing a code review, I would have a couple of comments here (including the one I've already mentioned about user-friendliness - the `update` calls - and responsibility-sharing problems (you have some checks which are performed by `tqdm` internally anyway)).
> 
> If you want to work on this problem, if I were you, I'd start by either finding or filing an issue with a detailed explanation of the problem you want to solve and code samples (preferably - using only standard libraries (like `threading`/`multiprocessing`), unless the problem only occurs with something specific like `numba`), then reading the current tqdm's source (at the tqdm/std.py file) and then making a PR.

Thanks for your reply.

rckoepke:
5116501f764e1b7447249359cb6ba09c05737fe0 worked perfectly for me in PyCharm on Ubuntu using "Emulate terminal in output console". Thank you!

kolayne:
@rckoepke, glad to hear that, thank you for your feedback!

Why did you use this commit, not the top of the branch? Does the behavior somehow worsen in the following commits or you just haven't tested?

rkbray:
To be honest, I independently found the same two-line solution and it was sufficient for my use-case. I went to submit a issue / PR for it and found this existing issue. It looked a bit stale and I so I wanted to provide additional positive feedback.

The other commits look very helpful and certainly better than the original two-line solution that I came up with, but I'm not sure how I would rigorously test them. I do have Windows+Mac+Linux and a variety of IDE's and terminal environments, so I'd be happy to do cross-platform testing if there was some kind of a test matrix that I could follow.

rckoepke:
To be honest, I independently found the same two-line solution and it was sufficient for my use case. I went to submit a issue / PR for it and found this existing PR that already had the same solution, plus apparently more comprehensive coverage. It looked a bit stale and I so I wanted to provide additional positive feedback.

The other commits look very helpful and certainly better than the original two-line solution that I came up with, but I'm not sure how I would rigorously test them. I do have Windows+Mac+Linux and a variety of IDE's and terminal environments, so I'd be happy to do cross-platform testing if there was some kind of a test matrix that I could follow.

casperdcl:
> [905453860](https://github.com/tqdm/tqdm/pull/1054#issuecomment-905453860): By the reaction you have left I can tell that you agree that the behavior found by espdev is incorrect

Oops, no that's not what I meant. Retracted the :+1:. What I mean is the output of [the code](https://github.com/tqdm/tqdm/pull/1054#issuecomment-787470013) is as expected. The `# it works incorrectly` comment should instead be `# auto-position it works as intended`, while the `# it works correctly` comment should be `# it doesn't work at all`.

Here's my output of auto-position (unaltered copy-paste of [this code](https://github.com/tqdm/tqdm/pull/1054#issuecomment-787470013)):

[asciicast](https://asciinema.org/a/Jv5lCEmmDvuAflsCgFp0PIoIg)

carlosgmartin:
@kolayne @casperdcl Came across this from #1000. Any update on this?

sigvaldm:
I also stumbled across this issue when using multiprocessing. The example in the README still does not work. Any update on when a solution might be merged into a stable release?

PR #1088, opened 2020-11-30
Fix formatting of '\n', '\0' in README

AlexeyDmitriev:
Another option is to use backicks


AlexeyDmitriev:
I doubt the CI problem is with my change.

PR #1189, opened 2021-06-19
telegram: add `delete()` when `leave=False`

raulsaavedr:
Add an option for delete bar message with the _**deleteMessage**_ method in Bot API.

### DIFF
--- a/tqdm/contrib/telegram.py
+++ b/tqdm/contrib/telegram.py
@@ -62,6 +62,19 @@ class TelegramIO(MonoWorker):
         else:
             return future
 
+    def delete(self):
+        try:
+            res = self.submit(
+                self.session.post,
+                self.API + '%s/deleteMessage' % self.token,
+                data=dict(
+                    chat_id=self.chat_id,
+                    message_id=self.message_id))
+        except Exception as e:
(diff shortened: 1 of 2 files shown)


raulsaavedr:
This would be the  behaviour

telegram_delete_method_pull

casperdcl:
Ideally this would be used when `leave=False`?

tqdm/contrib/telegram.py
```diff
@@ -112,6 +122,13 @@ def clear(self, *args, **kwargs):
         if not self.disable:
             self.tgio.write("")
 
+    def close(self):
+        if self.disable:
+            return
+        super(tqdm_telegram, self).close()
+        if not (self.leave or (self.leave is None and self.pos == 0)):
```
casperdcl, line 129: borrowing usual `leave=None|False|True` logic here.

Potential future issue is if people want different behaviour for terminal & telegram outputs (leave one but not the other). They'd have to override the `close()` method.

raulsaavedr: I think the leave variable is perfect, it fits our needs. And yes in the meantime we don't worry about that case.

PR #1190, opened 2021-06-21
logging: inherit output stream from existing handler

syzzer:
When using the logging redirection, logs will currently always be printed
to stdout, while the logging module default is to print to stderr.

Fix this by trying to inherit the stream from the existing handler, like
the code already does for the formatter.

Consider the following example:

    import logging
    import time

    from tqdm.contrib.logging import tqdm_logging_redirect

    log = logging.getLogger()
    log.warning("start")

    with tqdm_logging_redirect(range(int(4))) as pbar:
        for i in pbar:
            time.sleep(0.1)
            log.warning(f"Step {i}")

    log.warning("done")

Running this while redirecting stdout (`$ python3 log.py > /dev/null`)
without this patch will print:

    $ venv/bin/python log.py > /dev/null
    start
    100%|████████████████████████████████████████████| 4/4 [00:00<00:00,  9.87it/s]
    done

After this patch:

    $ venv/bin/python log.py > /dev/null
    start
    Step 0
    Step 1
    Step 2
    Step 3
    100%|████████████████████████████████████████████| 4/4 [00:00<00:00,  9.83it/s]
    done

### DIFF
--- a/tqdm/contrib/logging.py
+++ b/tqdm/contrib/logging.py
@@ -26,7 +26,7 @@ class _TqdmLoggingHandler(logging.StreamHandler):
     def emit(self, record):
         try:
             msg = self.format(record)
-            self.tqdm_class.write(msg)
+            self.tqdm_class.write(msg, file=self.stream)
             self.flush()
         except (KeyboardInterrupt, SystemExit):
             raise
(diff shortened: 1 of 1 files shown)


syzzer:
Sorry, this is incomplete. Don't waste any time reviewing yet please.

PR #1221, opened 2021-08-03
Avoid telegram limit hit

Hagai:
Using tqdm.contrib.telegram.tqdm with fast iteration results in "Too Many Requests" Error.

This issue is submitted in https://github.com/tqdm/tqdm/issues/1220.


Hagai:
I was referred to https://github.com/tqdm/tqdm/issues/1076#issuecomment-731147899 and saw I could use mininterval parameter to limit update rate. I've changed this pull request by setting the default mininterval for tqdm_telegram to 1 only for tqdm_telegram according to https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this.

Also, I've added a warning which detect "Too Many Requests" error upon TelegramIO initialization (It's critical to detect the error at this stage because it prevents setting the message_id attribute.

I think this pull request will provide more robustness to the telegram integration and better error reporting without any sacrifice

PR #1227, opened 2021-08-13
Test on Python 3.10 RC

hugovk:


hugovk:
Didn't mean to open this yet, will mark as draft until ready.

hugovk:
The workflow is being cancelled after 20 minutes, possibly a dependency is being built from source as it doesn't yet have a 3.10 wheel.

image

Will close for now and maybe check again later. Sorry for the noise!

PR #1295, opened 2022-01-25
Introducing tqdm.contrib.scp.ScpProgresswriter

Schwaneberg:
The scp library is handy for SSH/SFTP file operations, but it lacks a good console progress writer.
I'd like to contribute a class that provides a compatible progress writer using tqdm.
Please let me know what you think.


Schwaneberg:
It's not in progress, it is ready.
Please let me know if any changes, additional documentation or tests are desired.

casperdcl (changes requested):
Thanks!

Have you checked out the [keras](https://github.com/tqdm/tqdm/blob/master/tqdm/keras.py) implementation? If you could follow that here you can make use of inheritance, kwargs forwarding, `tqdm_auto` etc.

PR #1330, opened 2022-06-15
Update logging.py to deal with comet_ml

EmmaRenauld:
Hi,

When using comet in a project, the tqdm_logging_redirect fails (in duplicates messages). After searching, this comes from the
method tqdm.contrib.logging._is_console_logging_handler.

Without a comet environment, I get both sys.stderr and handler.stream as: `<_io.TextIOWrapper name='' mode='w' encoding='utf-8'>`.

However, with the comet environment, I get sys.stderr: `<comet_ml.console.Unbuffered object at 0x7fb7684e2af0>` but the handler.stream is still as before. 

Changing as proposed worked in my case, both when activating comet_ml or not. I hope this fix works in all cases.

Environment of test:
4.64.0 3.8.10 (default, Mar 15 2022, 12:22:08)
[GCC 9.4.0] linux

### DIFF
--- a/tqdm/contrib/logging.py
+++ b/tqdm/contrib/logging.py
@@ -36,7 +36,7 @@ class _TqdmLoggingHandler(logging.StreamHandler):
 
 def _is_console_logging_handler(handler):
     return (isinstance(handler, logging.StreamHandler)
-            and handler.stream in {sys.stdout, sys.stderr})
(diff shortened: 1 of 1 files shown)


EmmaRenauld:
Sorry, I discovered that when running pytest, the handler.stream (<_io.StringIO object) had no attribute name.

PR #1335, opened 2022-06-23
Right align n_fmt according to the total_fmt in the r_bar string

sindre0830:
While I used this library to create a progress bar for my PyTorch training loop, I found the values I had implemented in Prefix jumping around due to the ```n_fmt``` slowly increasing in width. This made it hard to read the data. 

I have added a right alignment to fix this and updated all tests to reflect these changes.

All tox tests passed.


sindre0830:
This is my first time contributing so let me know if there is anything I could have done differently.

Issue #1336, opened 2022-06-29
Visual statusbar finish error for multiprocessed tasks


MrJack91:
I use:
4.64.0 3.9.7 | packaged by conda-forge | (default, Sep 23 2021, 07:24:41) [MSC v.1916 64 bit (AMD64)] win32

While work is done everything is working:
![image](https://user-images.githubusercontent.com/219098/176372097-bfa0bf3a-de4f-4f19-aefa-3db6da5e0340.png)

But then every process which ends, wrotes his finish statusbar at the wrong position. This leads to:
![image](https://user-images.githubusercontent.com/219098/176371911-4c3df264-27b0-4339-a7f3-ad042b968731.png)

As far as i understand it is this line:
https://github.com/tqdm/tqdm/blob/master/tqdm/std.py#L1315-L1317

It should be:
```python
# let the finished status bar at this position it was
self.display(pos=pos)
fp_write('\r')
```

I started with:
```python
freeze_support()
tqdm.set_lock(RLock())
with Pool(initializer=tqdm.set_lock, initargs=(tqdm.get_lock(),), processes=8) as p:
    p.map(partial(Historizer._parallel), tasks)
```

Then i do:
```python
for obj in tqdm(object_set.fetchmany(100_000), desc=f'objs #{identifier:02d}', position=identifier):
```

MrJack91:
This would be the fix working in windows: https://github.com/tqdm/tqdm/pull/1337

mjpieters:
This is the same issue as #1496.

PR #1337, opened 2022-06-29
fix: windows wrong finish statusbar behaviour

MrJack91:
fix: windows wrong finish statusbar behaviour


MrJack91:
The corresponding issue: https://github.com/tqdm/tqdm/issues/1336

mjpieters:
This is not the right solution, unfortunately. Completed progress bars are explicitly pushed to the top to minimize updating (see #677 for why this happens). Your patch reverts this behaviour.

See #1502 for the right approach, which is to adjust the `pos` values of the remaining progress bars.

Issue #1342, opened 2022-07-11
Make tqdm(disable=None) default, instead of tqdm(disable=False)


sunyj:
- [x] I have marked all applicable categories:
    + [x] documentation request (i.e. "X is missing from the documentation." If instead I want to ask "how to use X?" I understand [StackOverflow#tqdm] is more appropriate)
    + [ ] new feature request
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [x] I have mentioned version numbers, operating system and
  environment, where applicable:
  ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)
  ```

From my personal experience (and many others), a progress bar is a great tool in developing but a nightmare for cronjob and logs.  It's hard (if not impossible) to find a case in which writing a progress bar to a log file can be justified as good, or even necessary.  I think it would be a wise choice to turn tqdm's output off to all non-TTY outputs BY DEFAULT.

janosh:
> From my personal experience (and many others), a progress bar is a great tool in developing but a nightmare for cronjob and logs. It's hard (if not impossible) to find a case in which writing a progress bar to a log file can be justified as good, or even necessary. I think it would be a wise choice to turn tqdm's output off to all non-TTY outputs BY DEFAULT.

Very much agree with this sentiment. Let's get #1343 merged.

yifanmai:
+1 for `disable=None` by default.

I noticed that this issue and the attached PR are stale. I'd be happy to help get this merged if needed.

asalahli:
While not the same, I was able to disable tqdm output globally by using `TQDM_DISABLE` environment variable - `TQDM_DISABLE=1 python script.py`

Couldn't find any documentation on it but it works thanks to this https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L951-L960

PR #1343, opened 2022-07-12
Enhancement: default disable to None, preventing output to log files by default.

sunyj:
Default disable to None.

Related issue: #1342


sunyj:
@casperdcl Would you please make a review and help get this PR merged?

djpohly:
@casperdcl This is a very small change that aligns tqdm's auto-enable behavior with that of other coloring/rich-text CLI utilities. Is there a concern that keeps this from being merged? It would be great to get it into a release. 👍🏼

PR #1350, opened 2022-08-08
Add optional message parameters to slack integration

JeronimoMendes:
Adds the slack sdk optional message parameters to the slack integration.

Related issue #1348


jdranczewski:
Hi! I've come to the repo planning to implement `thread_ts` as a pull request myself, so that `slack.contrib` can post in threads, but I see there's this pull request for it already. Any updates on whether this request looks like it will be merged or things I could help with to push it along?

JeronimoMendes:
Hey @jdranczewski! 
Some time has passed since this PR was opened and it never got a review. 
I'll tag @casperdcl and @0x2b3bfa0 (which was CCed on the original issue) in hopes it gets reviewed.
On your side maybe test this and see if works on your use case?
I'm open to making any changes required 😃

jdranczewski:
Can confirm this implementation is ok for my use case! I'm mostly interested in being able to start "sessions" where all progress bars go in a single thread, so I wrote a wrapper that does that by providing a `thread_ts` to your `slack_tqdm` and it works as intended:
```python
slack.tqdm(range(100), desc="No thread")
slack.tqdm.start_session("Testing threads")
slack.tqdm(range(100), desc="In thread")
slack.tqdm(range(100), desc="In thread")
```
(where `slack.tqdm` is my wrapper around `slack_tqdm`)

In the main channel:
slack_threads_1
In the tread:
slack_threads_2

A possible change suggestion would be to offload the optional Slack arguments into a `slack_kw` dictionary, similar to how matplotlib does `gridspec_kw` for the [Figure](https://matplotlib.org/stable/api/figure_api.html), just to avoid having all the additional arguments in the main `slack_tqdm.__init__`. One would then do:
```python
slack_tqdm(
    range(100),
    channel="dfghj", token="wertyu",
    slack_kw={
        "thread_ts": 12345,
        "icon_emoji": ":chart_with_upwards_trend:"
    }
)
```
I'm unsure whether that's actually cleaner in the long run, but it would decouple the tqdm API from having to keep up with possible changes to the Slack API (the `as_user` argument is already [considered "legacy"](https://api.slack.com/methods/chat.postMessage) for example). A different approach is considering if we need arguments other than `thread_ts`? `icon_emoji` doesn't seem to work in my testing, not sure if users need to access the parsing flags either since tqdm is providing most of the text in the message (see https://api.slack.com/methods/chat.postMessage for a full list).

JeronimoMendes:
I think it makes sense to add a `slack_kw` dict argument. Although the maintainer [didn't think of adding all slack arguments](https://github.com/tqdm/tqdm/issues/1348#issuecomment-1208058296) at the time, so not sure that will pass.

Sorry for the late reply.

PR #1351, opened 2022-08-10
Use dummy logger to provide hierarchal control of "disable" feature.

mrclary:
This builds off of #950 and adds the ability to control the disabled state of tqdm objects by setting the level of the associated dummy logger.

Dummy loggers are named `'tqdm' + '.' + desc`.
For example, given tqdm instances located anywhere:
- `tqdm(..., desc=None)`
- `tqdm(..., desc='a')`
- `tqdm(..., desc='a.b')`
- `tqdm(..., desc='a.c')`

all instances can be disabled with `logging.getLogger('tqdm').setLevel(30)`. The last three instances only can be disabled with `logging.getLogger('tqdm.a').setLevel(30)`. Or individual instances only can be disabled with `logging.getLogger('tqdm.a.b').setLevel(30)`.

Disabled states are set to True if the associated logger's effective level is >20, False otherwise.

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -11,6 +11,7 @@ from __future__ import absolute_import, division
 
 import os
 import sys
+import logging
 from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


mrclary:
Also, see comment: https://github.com/tqdm/tqdm/issues/614#issuecomment-1211048280

PR #1362, opened 2022-09-01
Keep logging level when using tqdm logger redirect

Serene-Arc:
This fixes an issue with the logging redirect. When using it, it removes the level of the associated logger that it is superseding, resulting in any verbosity setting being ignored from that point on. This is just a simple fix to carry over the logging level to the new logger object.


Serene-Arc:
@tfardet done! rebased off master. This isn't the only PR with these changes, fyi.

PR #1365, opened 2022-09-02
Model Train

pattonw:
tqdm is widely used in the machine learning community. It seems natural that we should be able to watch a model train while we watch our models train.
This pull request implements a new formatting option for the progress bar.
This code:
``` python
from tqdm import tqdm
import time


for _ in tqdm(range(1000), bar_format="{l_bar}{bar:t}{r_bar}"):
    time.sleep(0.1)
    pass
```

results in this progress bar:

[model_train-2022-09-02_13.55.33.webm](https://user-images.githubusercontent.com/14844826/188212880-0ec5764f-7b1b-46da-b0bb-dc85dd34ee57.webm)

To successfully watch a model train, you must have a nerd-font installed for the train, wagon, and tree characters.

I am completely unfamiliar with 99% of the tqdm code base, any advice on how to improve the model train or make it more customizable would be greatly appreciated.

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -205,9 +205,28 @@ class Bar(object):
         nsyms = len(charset) - 1
         bar_length, frac_bar_length = divmod(int(self.frac * N_BARS * nsyms), nsyms)
 
-        res = charset[-1] * bar_length
-        if bar_length < N_BARS:  # whitespace padding
-            res = res + charset[frac_bar_length] + charset[0] * (N_BARS - bar_length - 1)
+        import random
+
+        random.seed(1)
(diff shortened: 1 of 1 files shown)


jbsparrow:
@pattonw Please add an ssh or gpg signing certificate to your account because I really want this and the base branch requires all commits to be signed 😿

pattonw:
Thanks for the pointer. I added a signing key and went back to sign all my commits. Is that good? I know force pushing to replace commits isn't good practice but I couldn't find any other way to sign past commits

jbsparrow:
I think that should be good. Praying that a contributor approves it 🙏

Issue #1378, opened 2022-10-06
Rich tqdm progress reset fails because of missing argument


eblis:
- [x] I have marked all applicable categories:
    + [x] exception-raising bug
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [x] I have mentioned version numbers, operating system and environment, where applicable

There's seems to be a missing argument in the `reset` method definition:
```
    def reset(self, total=None):
        """
        Resets to 0 iterations for repeated use.

        Parameters
        ----------
        total  : int or float, optional. Total to use for the new bar.
        """
        if hasattr(self, '_prog'):
            self._prog.reset(total=total)
        super(tqdm_rich, self).reset(total=total)
```

glensc:
@eblis which argument is missing exactly? I'm comparing the source of `4.64.1` and I see the same code you pasted to the issue body.

PR #1379, opened 2022-10-06
Fix for tqdm.rich module, reset() method is broken - Added missing argument task_id to self._prog.reset() call

eblis:
There's seems to be a missing argument in the `reset` method definition for `tqdm.rich` module:
```python
    def reset(self, total=None):
        """
        Resets to 0 iterations for repeated use.

        Parameters
        ----------
        total  : int or float, optional. Total to use for the new bar.
        """
        if hasattr(self, '_prog'):
            self._prog.reset(total=total)
        super(tqdm_rich, self).reset(total=total)
```

I've created a fix for issue #1378 by adding the missing `task_id` parameter.
```python
self._prog.reset(task_id=self._task_id, total=total)
```

### DIFF
--- a/tqdm/rich.py
+++ b/tqdm/rich.py
@@ -139,7 +139,7 @@ class tqdm_rich(std_tqdm):  # pragma: no cover
         total  : int or float, optional. Total to use for the new bar.
         """
         if hasattr(self, '_prog'):
-            self._prog.reset(total=total)
+            self._prog.reset(task_id=self._task_id, total=total)
         super(tqdm_rich, self).reset(total=total)
 
 


glensc:
@eblis noticed you have submitted the fix. please fix pull request title to indicate you are fixing `tqdm.rich` module. and describe the problem in the body (copy it).

eblis:
I updated the title and first comment.
I see an error about signing commits, do you want me to sign my commit as well ? I don't have that configured in github (yet), will need some time to configure.

glensc:
I don't see signin error. i'm sure it's not relevant.

glensc (approved):

PR #1384, opened 2022-10-20
Clarify licence in package metadata

RazerM:
Some packages provide their code under multiple licences (e.g. MIT OR Apache-2.0), while the tqdm/LICENCE file indicates that there are contributions given under separate licences, resulting in a licence of MPL-2.0 AND MIT.

The goal of this change is to help downstream users to disambiguate cases where multiple licences are involved. Although PEP 639 is not yet here, nothing stops individual packages from providing a SPDX license expression in the existing license metadata field, which is what I've chosen here.

## Additional information

<details>
<summary>Parsed licence expression example</summary>

```python
In [1]: from license_expression import ExpressionError, get_spdx_licensing

In [2]: licensing = get_spdx_licensing()

In [3]: licensing.parse('MPL-2.0 AND MIT', strict=True, validate=True)
Out[3]: AND(LicenseSymbol('MPL-2.0', is_exception=False), LicenseSymbol('MIT', is_exception=False))
```

</details>

[SPDX License List](https://spdx.org/licenses/)


YasinAshorian (approved):

RazerM:
I see this was implemented in the migration to pyproject.toml, thanks!

PR #1473, opened 2023-05-12
contrib.concurrent: fix iterable length handling

svank:
`contrib.concurrent` has functions for running `map`-like functions in parallel with a progress bar. This PR makes three fixes to how the lengths of the iterables are detected.

If the function provided to `process_map` or `thread_map` accepts `N` arguments, these map functions accept `N` iterables, with each iterable providing values for one argument. The map function keeps calling the provided function until reaching the end of the shortest iterable.

`process_map` raises a warning if a long iterable has been provided but a `chunksize` has not been set. It was looking at the length of the longest iterable, but it should look at the shortest length, as that's what sets the number of iterations.

Prior behavior:
```python
In [7]: from tqdm.contrib.concurrent import process_map
   ...: from time import sleep
   ...: from random import random
   ...: def f(*args):
   ...:     sleep(random() * 3)
   ...: process_map(f, range(2000), range(2))
<ipython-input-7-cc99b55bb53a>:6: TqdmWarning: Iterable length 2000 > 1000 but `chunksize` is not set. This may seriously degrade multiprocess performance. Set `chunksize=1` or more.
  process_map(f, range(2000), range(2))
  0%|                                          | 2/2000 [00:01<26:27,  1.26it/s]
Out[7]: [None, None]
```
Note the warning about a length-2000 iterable, despite only running two iterations. With this PR, the warning is not raised in this instance.

Note also that the progress bar shows an expected total of 2000 iterations, despite ending after 2. This is because `_executor_map` was only looking at the first iterable to set the expected total. This PR changes this behavior to use the length of the shortest iterable.

New behavior:
```python
In [1]: from tqdm.contrib.concurrent import process_map
   ...: from time import sleep
   ...: from random import random
   ...: def f(*args):
   ...:     sleep(random() * 3)
   ...: process_map(f, range(2000), range(2))
100%|█████████████████████████████████████████████| 2/2 [00:02<00:00,  1.23s/it]
Out[1]: [None, None]
```

This PR also improves how these map functions handle iterables with no length. (One pattern would be having one variable and one fixed argument, and calling `process_map(my_function, range(10), itertools.repeat(fixed_value))`.) Because of the default length of zero returned by `operator.length_hint` (and with this PR's emphasis on the *shortest* length), no-length iterables caused the expected total to be 0. The second commit in this PR ignores those lengths when calculating the expected total.

Before:
```python
In [20]: from tqdm.contrib.concurrent import process_map
    ...: from time import sleep
    ...: from random import random
   ...: from itertools import repeat
    ...: def f(*args):
    ...:     sleep(random() * 3)
    ...: process_map(f, range(2000), range(2), repeat(1))
2it [00:02,  1.11s/it]
Out[20]: [None, None]
```

After:
```python
In [2]: from tqdm.contrib.concurrent import process_map
   ...: from time import sleep
   ...: from random import random
   ...: from itertools import repeat
   ...: def f(*args):
   ...:     sleep(random() * 3)
   ...: process_map(f, range(2000), range(2), repeat(1))
100%|█████████████████████████████████████████████| 2/2 [00:01<00:00,  1.95it/s]
Out[2]: [None, None]
```

### DIFF
--- a/tests/tests_concurrent.py
+++ b/tests/tests_concurrent.py
@@ -38,7 +38,9 @@ def test_process_map():
 
 @mark.parametrize("iterables,should_warn", [([], False), (['x'], False), ([()], False),
                                             (['x', ()], False), (['x' * 1001], True),
-                                            (['x' * 100, ('x',) * 1001], True)])
+                                            (['x' * 100, ('x',) * 1001], False),
+                                            (['x' * 1001, ('x',) * 100], False),
+                                            (['x' * 1001, ('x',) * 1001], True)])
 def test_chunksize_warning(iterables, should_warn):
     """Test contrib.concurrent.process_map chunksize warnings"""
     patch = importorskip('unittest.mock').patch
--- a/tqdm/contrib/concurrent.py
+++ b/tqdm/contrib/concurrent.py
@@ -39,7 +39,8 @@ def _executor_map(PoolExecutor, fn, *iterables, **tqdm_kwargs):
     """
     kwargs = tqdm_kwargs.copy()
     if "total" not in kwargs:
-        kwargs["total"] = length_hint(iterables[0])
+        shortest_iterable_len = min(map(length_hint, iterables))
+        kwargs["total"] = shortest_iterable_len
     tqdm_class = kwargs.pop("tqdm_class", tqdm_auto)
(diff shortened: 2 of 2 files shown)


casperdcl (approved):
Apologies for the delay!

PR #1492, opened 2023-08-08
build: do not install `examples` package

mgorny:
Fix `pyproject.toml` to exclude the `examples` directory from being installed as a top-level package.


casperdcl (approved):

mgorny:
Thanks!

PR #1522, opened 2023-10-13
Add support for enumerate

vincentblot28:
Enable the use of "enumerate" as iterable of tqdm. With this feature, tqdm automatically compute the total number of iterations when the argument of tqdm is of type "enumerate"

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -976,7 +976,10 @@ class tqdm(Comparable):
 
         if total is None and iterable is not None:
             try:
-                total = len(iterable)
+                if isinstance(iterable, enumerate):
+                    total = len(list(iterable))
+                else:
+                    total = len(iterable)
             except (TypeError, AttributeError):
                 total = None
         if total == float("inf"):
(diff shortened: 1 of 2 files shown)


tqdm/std.py
```diff
@@ -976,7 +976,10 @@ def __init__(self, iterable=None, desc=None, total=None, leave=True, file=None,
 
         if total is None and iterable is not None:
             try:
-                total = len(iterable)
+                if isinstance(iterable, enumerate):
+                    total = len(list(iterable))
+                else:
+                    total = len(iterable)
```
niqodea, line 982: I am not sure this change will be approved. By calling `list` on the `enumerate` iterable you are potentially loading an entire dataset into memory.

niqodea: You could extract the wrapped iterable and work with that instead.

niqodea: Anyhow, the standard way to use `tqdm` together with `enumerate` is to wrap with `tqdm` first and `enumerate` second:

```python
enumerate(tqdm(iterable))
```

PR #1535, opened 2023-11-28
Replace deprecated disco-py client in tqdm.contrib.discord by discord.py

guigoruiz1:
This PR is an extension to PR #1383 with the capability of choosing which `channel_id` to post the progress bar maintained. In this PR the user can choose between the simplified webhook method used in PR #1383 or using a `token` and `channel_id` as done before with disco-py. The handling of the client when using token and channel_id is asynchronous and relies on `asyncio`.  I have tested all variations and it works, although the webhook method is slightly faster to start.

### DIFF
--- a/tqdm/contrib/discord.py
+++ b/tqdm/contrib/discord.py
@@ -125,8 +125,8 @@ class tqdm_discord(tqdm_auto):
                     )
             else:
                 self.dio = DiscordIO(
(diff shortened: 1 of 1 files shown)


guigoruiz1:
I found a bug which required asking `nest_asyncio` to fix.

guigoruiz1:
Closing as this solution is too cumbersome. Will open another PR with the simplest working solution I found.

PR #1530, opened 2023-11-07
Fix docstring of thread_map

albertvillanova:
Replace `max` with `min` in `contrib.concurrent.thread_map` docstring about default `max_workers`.

PR #1543, opened 2023-12-28
contrib.concurrent: Use python default max_workers

aripollak:
I noticed that contrib.concurrent defines its own default for max_workers, which is the same as CPython's default for [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor) in CPython 3.8+, but differs from the default for [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor):
> If max_workers is None or not given, it will default to the number of processors on the machine...If max_workers is None, then the default chosen will be at most 61, even if more processors are available.

This was surprising to me. If the running code is CPU-bound, which is usually why you'd use `process_map` instead of `thread_map`, there usually isn't a noticeable speed improvement to spawn more processes than the number of available CPUs. In my case, `cpu_count + 4` ended up using more RAM than necessary for maximum speed improvement. I'm using a package that loads an index into memory at 800 MB per process, so that's an unnecessary extra 3.2 GB used by default.

Obsoletes #1530 and #1518.

### DIFF
--- a/tqdm/contrib/concurrent.py
+++ b/tqdm/contrib/concurrent.py
@@ -3,7 +3,6 @@ Thin wrappers around `concurrent.futures`.
 """
 from contextlib import contextmanager
 from operator import length_hint
-from os import cpu_count
 
 from ..auto import tqdm as tqdm_auto
 from ..std import TqdmWarning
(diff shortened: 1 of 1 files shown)


casperdcl (approved):
Apologies for the delay!

PR #1552, opened 2024-02-11
add keyword 'repeat' for tqdm.contrib.itertools.product

unickname3:
At now keyword 'repeat' don't work in tqdm.contrib.itertools.product()

From itertools docs:

> itertools.product(*iterables, repeat=1)
> ....
> To compute the product of an iterable with itself, specify the number of repetitions with the optional repeat keyword argument. For example, product(A, repeat=4) means the same as product(A, A, A, A).

There is my solution for this problem.

### DIFF
--- a/tqdm/contrib/itertools.py
+++ b/tqdm/contrib/itertools.py
@@ -9,7 +9,7 @@ __author__ = {"github.com/": ["casperdcl"]}
 __all__ = ['product']
 
 
-def product(*iterables, **tqdm_kwargs):
+def product(*iterables, repeat=1, **tqdm_kwargs):
     """
     Equivalent of `itertools.product`.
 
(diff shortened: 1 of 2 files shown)


bartbroere:
Hi, I made a similar PR some time ago: #1428. It's still open.

I don't care really which one of these gets merged, but now they're both linked at least.

tqdm/contrib/itertools.py
```diff
@@ -29,7 +29,7 @@ def product(*iterables, **tqdm_kwargs):
             total *= i
```
bartbroere, line 29: Apart from some minor differences (in the tests for example), I only found one difference that seems to matter. I believe that the influence of actually using the `repeat` parameter is not yet reflected in the `total` in your version. On my branch I do this to get a correct `total`:

```suggestion
            total *= i
        total = total ** repeat
```

Without this fix I believe the progress bar will exceed 100% (after which it just starts displaying the amount of iterations).

PR #1572, opened 2024-04-25
Add support for setting mininterval and maxinternal via environment variables

eendebakpt:
Add support for setting mininterval and maxinternal via environment variables


casperdcl (commented):
this shouldn't be needed as the default values (`mininterval=0.1, maxinterval=10.0`) are already used to infer the `float` type.

Happy to re-open if this is still a problem.

PR #1576, opened 2024-04-29
contrib.concurrent: timeout, buffersize

shermansiu:
As mentioned in https://docs.python.org/3/library/concurrent.futures.html, ThreadPoolExecutor.map and ProcessPoolExecutor.map both accept the timeout argument, which is not currently offered in thread_map and process_map.

### DIFF
--- a/tqdm/contrib/concurrent.py
+++ b/tqdm/contrib/concurrent.py
@@ -1,7 +1,6 @@
 """
 Thin wrappers around `concurrent.futures`.
 """
-
 from contextlib import contextmanager
 from operator import length_hint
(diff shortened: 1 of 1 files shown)


shermansiu:
@casperdcl @lrq3000 I've noticed that there's a huge backlog of PRs that haven't been reviewed in a while. There is also a huge gap in time between when PRs are submitted and when they are reviewed. 

Is there a need to onboard some maintainers or is `tqdm` in maintenance mode?

casperdcl (approved):
Apologies for the delay!

Also added `buffersize`.

Issue #1619, opened 2024-09-29
Keras TqdmCallback handles poorly models with many metrics


LucaCappelletti94:
## Issue description

At this time, when a keras model has many outputs and/or many metrics being tracked, the loading bar tries to display them all by default with poor outcomes: necessarily, not all features are displayed, and due to the many features being tentatively displayed the loading bar breaks into multiple lines, as shown in the attached figure. In this specific screenshot, there is also the Keras `ModelCheckpoint` spamming between lines, but even without that the issue persists.

<img width="1792" alt="Screenshot 2024-09-29 alle 10 03 58" src="https://github.com/user-attachments/assets/d8291093-7ebd-4d5f-ad1a-36767671ca7b">

### Proposed solution

Since not all metrics can be reasonably shown at once, I wanted to add an argument to let the user specify by name which metrics to be displayed. If you agree, I can go ahead with the pull request. I had in mind something of this nature:

```python
TqdmCallback(
     verbose=2,
     metrics=["val_loss", "loss"]
)
```

## Issue form

- [x] I have marked all applicable categories:
    + [ ] documentation request (i.e. "X is missing from the documentation." If instead, I want to ask "how to use X?" I understand [StackOverflow#tqdm] is more appropriate)
    + [x] new feature request
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [ ] I have mentioned version numbers, operating system and
  environment, where applicable:
  ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)
  ```
  
  `4.66.5 3.12.5 | packaged by Anaconda, Inc. | (main, Sep 12 2024, 18:27:27) [GCC 11.2.0] Linux`

[source website]: https://github.com/tqdm/tqdm/
[known issues]: https://github.com/tqdm/tqdm/#faq-and-known-issues
[issue tracker]: https://github.com/tqdm/tqdm/issues?q=
[StackOverflow#tqdm]: https://stackoverflow.com/questions/tagged/tqdm

PR #1620, opened 2024-09-29
Added support for 'metrics' parameter in Keras TQDM Callback

LucaCappelletti94:
This pull request resolves issue #1619 by introducing a `metrics` keyword argument to specify which metrics should be displayed. Furthermore, it resolves code smells reported by Pylint within the TqdmCallback document.


LucaCappelletti94:
It has failed with `Failed to pull source: 'source-commit' cannot be used with a local source.` - I am not sure what that means

PR #1628, opened 2024-11-04
Improve robustness of IPykernel detection

Carreau:
This is not perfect (it's technically impossible to really know) but should help; in particular it is possible to launch ipykernel without `'IPykernel'` being present in the config, in the case of cloud based launcher (like
https://gateway-provisioners.readthedocs.io/en/latest/).

Note that with the ubiquitousness of tqdm, if this makes it easier for you; I'm open to add a

```
try:
    import tqdm
    tqdm.in_ipykernel(...)
except ImportError:
    pass
```

Or anything similar if this helps detecting in which frontend you are run from. I'm guessing this coudl help as sometime ipykernel is in a notebook, but sometime in QT (like in spyder), and I guess it could help decide what to do without user interactions.

### DIFF
--- a/tqdm/autonotebook.py
+++ b/tqdm/autonotebook.py
@@ -10,8 +10,14 @@ import sys
 from warnings import warn
 
 try:
+    if 'ipykernel' not in sys.modules:
+        raise ImportError("console")
     get_ipython = sys.modules['IPython'].get_ipython
-    if 'IPKernelApp' not in get_ipython().config:  # pragma: no cover
+    ipython = get_ipython()
+    if not ipython:
+        raise ImportError("console")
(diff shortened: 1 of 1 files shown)


casperdcl (approved):
Apologies for the delay; lgtm

Carreau:
I've rebased on main,

Let me know if there is anything else I can do.

Carreau:
Poking at this again

casperdcl:
Having finally tested various Jupyter frontends locally (and found the StackOverflow suggestions of checking `ZMQInteractiveShell` are no longer valid), this should now be fixed in [`tqdm>=4.68.3`](https://tqdm.github.io/releases/#v4683-2026-06-17) by #1763!

Carreau:
thanks

PR #1649, opened 2025-01-27
std: release lock on exception in clear/refresh

skshetry:


skshetry:
Hey @casperdcl. I hope you're doing well. Just wanted to ask if you could take a look at this when you have some time. Thanks.

PR #1669, opened 2025-06-10
add name for the monitoring thread

FT9R:
Added name for monitoring thread for tqdm bars.

Before:
image

After:
image


casperdcl:
apologies for the delay; fixed in #1752

PR #1596, opened 2024-07-14
Update `tqdm.rich`

klamike:
To try it out for yourself, use:
```bash
pip install --force-reinstall git+https://github.com/klamike/tqdm.git@mk/update_tqdmrich
```

___
This PR updates `tqdm.rich` to better match `tqdm.std`. It's a lot packed into one PR, but I hope you will still consider merging as it significantly extends the current functionality.

- use `tqdm.std.format_*` instead of `rich.filesize` for unit number formatting
- support iterators without` __len__`
- add inverse speed support
- add `NoPaddingProgress` & `ProgressColumn` classes to match `tqdm.std` style
- add `set_postfix` support (#1510)
- add `ascii` support
- use only the lock in rich -- avoid the tqdm lock
- add `bar_options` kwarg to customize `BarColumn`
- support `colour` kwarg on init
- support simultaneous `tqdm.rich` bars via global `cls._progress`
- use lazy auto refresh from `rich`
- support `dynamic_ncols`
- refresh on last display (#1580)

Closes #1674
Closes #1644

PR #1681, opened 2025-08-13
feat: EMA speed and remaining time for tqdm_rich

beskep:
The speed and remaining time of `tqdm_rich`, which were calculated as the overall average, were replaced with the EMA calculation of `tqdm`.


beskep:
Closing as #1596 already fixed it.

PR #1692, opened 2025-11-16
Use binary SI prefixes if unit_divisor is 1024

native-api:
Subj.

Addresses https://github.com/tqdm/tqdm/issues/354#issuecomment-652886495


eliotwrobson:
@native-api I have implemented this feature on my fork https://github.com/eliotwrobson/tldm

Issue #1675, opened 2025-07-25
Clarify supported Python versions in installation docs


Sodas0:
- [x] I have marked all applicable categories:
    + [x] documentation request (i.e. "X is missing from the documentation." If instead I want to ask "how to use X?" I understand [StackOverflow#tqdm] is more appropriate)
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [x] I have mentioned version numbers, operating system and
  environment, where applicable:
 ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)
 ```

[source website]: https://github.com/tqdm/tqdm/
[known issues]: https://github.com/tqdm/tqdm/#faq-and-known-issues
[issue tracker]: https://github.com/tqdm/tqdm/issues?q=
[StackOverflow#tqdm]: https://stackoverflow.com/questions/tagged/tqdm

----

The installation instructions in the README (https://github.com/tqdm/tqdm#installation) do not mention the supported Python versions for tqdm. This would be helpful for users with older environments or when debugging version compatibility.

I suggest adding a line such as: "tqdm officially supports Python 3.7 and above."

Let me know if a PR for this would be welcome.

yajie628:
Hi, I would like to help with this issue.

I have checked the README and the docs installation section and confirmed that the supported Python versions are not explicitly stated.

My plan:
- Verify the actual supported Python versions from `setup.cfg` and PyPI metadata
- Update the installation docs / README to include a clear statement (e.g., "tqdm officially supports Python 3.7+")
- Submit a documentation-only PR

Please let me know if this approach sounds good. Thanks!

PR #1694, opened 2025-11-29
Clarify supported Python versions in installation docs

yajie628:
This PR improves the installation documentation by explicitly stating the
supported Python versions for tqdm.

### What was the issue?
The README and installation instructions did not clearly mention the
minimum Python version supported by tqdm. This could cause confusion for users on
older Python environments.

Related issue: https://github.com/tqdm/tqdm/issues/1675

### What was changed?
- Added a line in `README.rst` under the Installation section:
  > tqdm officially supports Python 3.7 and above.

### Why is this useful?
- Improves clarity for new users
- Aligns documentation with the `python_requires >=3.7` specified in `setup.cfg`
- Reduces compatibility-related questions from users

### Notes
This PR modifies documentation only and does not affect code behavior.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,7 @@ Python and an environment supporting ``carriage return \r`` and
 
 Installation
 ------------
+tqdm officially supports Python 3.7 and above.
 
 Latest PyPI stable release
 ~~~~~~~~~~~~~~~~~~~~~~~~~~


mxr:
IMO, it would be better to say something like "the minimum version is in the `requires-python` key of `pyproject.toml`" so you don't need to update the README every time this minimum changes

PR #1707, opened 2026-02-08
fix: pass unit_divisor to format_sizeof when formatting rate

### LINKED ISSUE #1690: The value of unit_divisor does not affect the calculation of rate_fmt.
```python
from tqdm import tqdm
import time

for i in tqdm(range(10000),
              unit="B",
              unit_scale=True,
              unit_divisor=1024,
              bar_format="{l_bar}{bar:100}| {n_fmt}/{total_fmt} [{elapsed}<{remaining}, {rate_fmt} {rate}]",
              ):
    time.sleep(0.00025)
```

Image

I set the value of unit_divisor to 1024, so rate_fmt should also use this as the scaling factor, but why is it still using 1000 as the scaling factor? How can I fix this issue?

veeceey:
## Summary
- When `unit_scale=True` and a custom `unit_divisor` is set (e.g., 1024 for binary units), `rate_fmt` and `rate_inv_fmt` were not respecting the `unit_divisor` value
- The `format_sizeof` calls for rate formatting were missing the `divisor=unit_divisor` parameter, causing them to always use the default divisor of 1000
- This made the rate display inconsistent with `n_fmt` and `total_fmt`, which correctly passed `unit_divisor`

Fixes #1690

## Test plan
- All 71 existing tests pass
- Verified that `format_meter(n=5000, total=10000, elapsed=10, unit='B', unit_scale=True, unit_divisor=1024, rate=2048)` now correctly shows `2.00kB/s` (using 1024 divisor) instead of the previous incorrect `2.05kB/s` (using 1000 divisor)

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -551,10 +551,10 @@ class tqdm(Comparable):
             rate = (n - initial) / elapsed
         inv_rate = 1 / rate if rate else None
         format_sizeof = tqdm.format_sizeof
-        rate_noinv_fmt = ((format_sizeof(rate) if unit_scale else f'{rate:5.2f}')
+        rate_noinv_fmt = ((format_sizeof(rate, divisor=unit_divisor) if unit_scale else f'{rate:5.2f}')
                           if rate else '?') + unit + '/s'
         rate_inv_fmt = (
-            (format_sizeof(inv_rate) if unit_scale else f'{inv_rate:5.2f}')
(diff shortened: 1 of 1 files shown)


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

veeceey:
hey, any chance this could get reviewed? the unit_divisor was being ignored for rate formatting which gives wrong output

eliotwrobson:
@veeceey it looks like this repo isn't regularly maintained, but I have a fork here which I think has this fix already https://github.com/eliotwrobson/tldm

PR #1714, opened 2026-02-22
Support async iterables with only __aiter__ in tqdm_asyncio

### LINKED ISSUE #1686: tqdm_asyncio doesn't support all async iterables
`tqdm_asyncio` supports async iterables with `__anext__` but not those with only `__aiter__` such as [`aiomultiprocess.PoolResult`
](https://github.com/omnilib/aiomultiprocess/blob/main/aiomultiprocess/pool.py#L121-L145)

bysiber:
Fixes #1686.

`tqdm_asyncio.__init__` checks for `__anext__` (async iterator protocol) and `__next__` (sync iterator), but doesn't handle objects that only have `__aiter__` without `__anext__` (async iterable protocol). These objects (like `aiomultiprocess.PoolResult`) fall through to `iter()` which doesn't work on async-only iterables.

Added a check for `__aiter__` between the existing `__anext__` and `__next__` checks. When an object only has `__aiter__`, we call `__aiter__()` to get the actual async iterator, then use its `__anext__` method. This follows the same pattern as the existing `else` branch that calls `iter()` for sync iterables that only have `__iter__`.

Added a test with a custom async iterable class that only provides `__aiter__`.

### DIFF
--- a/commit_msg.txt
+++ b/commit_msg.txt
@@ -1,0 +1,9 @@
+Support async iterables with only __aiter__ in tqdm_asyncio
+
+tqdm_asyncio only checked for __anext__ (async iterator) and __next__
+(sync iterator), but did not handle objects that only implement
+__aiter__ (async iterable). Such objects would fall through to the
+sync iter() call and fail.
+
+Added a check for __aiter__ between the __anext__ and __next__ checks,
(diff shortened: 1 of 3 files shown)


casperdcl (approved):
Thanks! Apologies for the delay.

PR #1720, opened 2026-02-27
fix: tqdm_notebook.update() no longer returns True

### LINKED ISSUE #1716: tqdm_notebook.update() returns True, causing unwanted output in Jupyter
## Description

`tqdm.update()` returns `True` when a display refresh is triggered (documented behavior). However, in Jupyter notebooks, this return value gets printed by the REPL when `update()` is the last expression in a loop body or cell, producing unwanted `True` output — potentially hundreds of times in long loops.

## Minimal reproducer

```python
from tqdm.auto import tqdm
from concurrent.futures import ThreadPoolExecutor, as_completed

pbar = tqdm(total=100)
for i in range(100):
    pbar.update()  # prints "True" ~100 times in Jupyter
pbar.close()
```

## Current workaround

```python
_ = pbar.update()
```

## Proposed fix

The notebook subclass in `tqdm/notebook.py` currently passes through the return value:

```python
def update(self, n=1):
    try:
        return super().update(n=n)
    except:
        self.disp(bar_style='danger')
        raise
```

Since notebook users are the ones affected by Jupyter's auto-display of last-expression return values, `tqdm_notebook.update()` could simply not return the result:

```python
def update(self, n=1):
    try:
        super().update(n=n)
    except:
        self.disp(bar_style='danger')
        raise
```

This preserves the return value in `tqdm.std.update()` for programmatic use, while preventing the notebook-specific footgun. Alternatively, `tqdm.std.update()` itself could return `None` — the `True` sentinel is rarely used by callers in practice.

## Environment

- tqdm 4.x (all recent versions)
- Jupyter Notebook / JupyterLab
- Python 3.10+

worksbyfriday:
## Summary

- `tqdm_notebook.update()` now returns `None` instead of passing through the base class `True`
- Prevents Jupyter's REPL from auto-displaying `True` hundreds of times when `pbar.update()` is the last expression in a loop body
- The base `std.tqdm.update()` still returns `True`/`None` for programmatic use

## Reproducer (before fix)

```python
from tqdm.auto import tqdm

pbar = tqdm(total=100)
for i in range(100):
    pbar.update()  # prints "True" ~100 times in Jupyter
pbar.close()
```

## Change

One-line change in `tqdm/notebook.py`: remove `return` from `return super().update(n=n)` so the notebook subclass doesn't pass through the return value.

## Test

Added `test_notebook_update_returns_none` to verify the notebook subclass returns `None`.

Fixes #1716

> I'm an AI (Claude, worksbyfriday). All code is my own — reviewed and tested before submission.

### DIFF
--- a/tests/tests_notebook.py
+++ b/tests/tests_notebook.py
@@ -5,3 +5,10 @@ def test_notebook_disabled_description():
     """Test that set_description works for disabled tqdm_notebook"""
     with tqdm_notebook(1, disable=True) as t:
         t.set_description("description")
+
+
+def test_notebook_update_returns_none():
+    """Test that tqdm_notebook.update() returns None to avoid Jupyter auto-display"""
+    with tqdm_notebook(total=10, disable=True) as t:
+        result = t.update(1)
+        assert result is None
--- a/tqdm/notebook.py
+++ b/tqdm/notebook.py
@@ -257,7 +257,7 @@ class tqdm_notebook(std_tqdm):
 
     def update(self, n=1):
         try:
-            return super().update(n=n)
+            super().update(n=n)
         # NB: except ... [ as ...] breaks IPython async KeyboardInterrupt
         except:  # NOQA
             # cannot catch KeyboardInterrupt when using manual tqdm


casperdcl:
As per https://tqdm.github.io/docs/tqdm/#update, the `update()` deliberately returns a `bool` (which can be useful for users wanting to piggy back off `tqdm`'s intelligent anti-console-spamming logic to do some conditional processing).

The correct way to suppress final-cell-line output in jupyter is to [suffix a `;`](https://ipython.readthedocs.io/en/stable/interactive/tips.html#suppress-output) or to [prefix `%%capture`](https://ipython.readthedocs.io/en/stable/interactive/magics.html#cellmagic-capture) to your cell.

PR #1724, opened 2026-03-19
Add flush argument to tqdm.write

### LINKED ISSUE #1651: Supporting `flush` arg in `tqdm.write` to make drop-in replacement for `print`
I am trying to move from `print` to `tqdm.write`. However, some of my `print` have `flush=True`, which is an argument not present in `tqdm.write`.

The request is, can `tqdm.write` have a `flush` argument added?

Searching "flush" in https://tqdm.github.io/, there are no results.

---

```pycon
4.67.1 3.12.8 (main, Jan 28 2025, 10:06:03) [Clang 16.0.0 (clang-1600.0.26.4)] darwin
```

KGuptaLive:
Closes #1651

## Summary

Adds a `flush=False` keyword argument to `tqdm.write()` to match the
signature of Python's built-in `print(flush=True)`.

When `flush=True`, `fp.flush()` is called after writing.

## Changes

- `tqdm/std.py` — `flush=False` added to `write()` signature; `fp.flush()` called when `True`
- `tests/tests_tqdm.py` — `test_write_flush()` added

## Test plan

- `test_write_flush` passes
- 130 passed, 7 skipped (optional deps: numpy, dask, keras, rich, tkinter)

### DIFF
--- a/tests/tests_tqdm.py
+++ b/tests/tests_tqdm.py
@@ -6,6 +6,7 @@ import re
 import sys
 from contextlib import contextmanager
 from functools import wraps
+from unittest.mock import MagicMock, call
 from warnings import catch_warnings, simplefilter
 
 from pytest import importorskip, mark, raises, skip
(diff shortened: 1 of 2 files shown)


eliotwrobson:
@KGuptaLive this feature is implemented on my fork https://github.com/eliotwrobson/tldm

PR #1725, opened 2026-03-23
fix(notebook): suppress update() return to prevent unwanted True in Jupyter (#1716)

### LINKED ISSUE #1716: tqdm_notebook.update() returns True, causing unwanted output in Jupyter

r266-tech:
## Summary

Fixes #1716

In Jupyter notebooks, the REPL auto-displays the return value of the last expression in a cell. `tqdm.update()` returns `True` when a display refresh is triggered, causing unwanted `True` output — potentially hundreds of times in long loops.

## Changes

- Removed the `return` statement from `tqdm_notebook.update()`, keeping the `super().update()` call intact.

## Why this approach

- The `True` return value from `update()` is a sentinel indicating a display refresh was triggered. This is rarely used by callers and never useful in notebook contexts.
- This change only affects the notebook subclass, leaving `tqdm.std.update()` untouched for programmatic use.
- Minimal, one-line change with no behavioral side effects beyond suppressing the display of the return value.

## Reproducer (before fix)

```python
from tqdm.auto import tqdm
pbar = tqdm(total=100)
for i in range(100):
    pbar.update()  # prints "True" ~100 times in Jupyter
pbar.close()
```

## Testing

The change is a one-line removal of `return` from an already-existing try/except block. The exception handling path (`self.disp(bar_style='danger'); raise`) is unaffected.

### DIFF
--- a/tqdm/notebook.py
+++ b/tqdm/notebook.py
@@ -257,7 +257,7 @@ class tqdm_notebook(std_tqdm):
 
     def update(self, n=1):
         try:
-            return super().update(n=n)
+            super().update(n=n)
         # NB: except ... [ as ...] breaks IPython async KeyboardInterrupt
         except:  # NOQA
             # cannot catch KeyboardInterrupt when using manual tqdm


casperdcl:
Duplicate of #1720.

Also, as per https://tqdm.github.io/docs/tqdm/#update, the `update()` deliberately returns a `bool` (which can be useful for users wanting to piggy back off `tqdm`'s intelligent anti-console-spamming logic to do some conditional processing).

The correct way to suppress final-cell-line output in jupyter is to [suffix a `;`](https://ipython.readthedocs.io/en/stable/interactive/tips.html#suppress-output) or to [prefix `%%capture`](https://ipython.readthedocs.io/en/stable/interactive/magics.html#cellmagic-capture) to your cell.

PR #1736, opened 2026-04-04
Add optional ETA smoothing for stable streaming estimates

Steve2457:
### Summary

- Adds a new optional parameter `eta_smoothing` to reduce ETA jitter in streaming/download workloads with unstable throughput.
- Keeps backward compatibility: default is `eta_smoothing=0`, which preserves existing behavior.
- Applies smoothing at the ETA-display layer (after `current_eta = (total - n) / rate`), without changing iteration counting or update cadence.
- Uses EMA to smooth `remaining_s`/`eta` output so ETA is less jumpy under bursty network conditions.
- Resets ETA smoothing state on `reset()` to avoid stale state across reused bars.
- Updates docs to include the new option in both README and manpage.
- Adds regression coverage for ETA volatility reduction with deterministic timing patterns.

### Test plan

- [x] Run targeted regression test for new behavior:
  - `python -m pytest tests_tqdm.py -k test_eta_smoothing -v`
- [x] Run full tqdm core test file to ensure no regressions:
  - `python -m pytest tests/tests_tqdm.py`
- [x] Verify default behavior is unchanged (`eta_smoothing=0`):
  - Existing ETA and smoothing tests still pass.
- [x] Verify smoothing effect is measurable when enabled:
  - `test_eta_smoothing` asserts reduced peak and total ETA volatility for `eta_smoothing>0`.
- [x] Manual smoke check with unstable-delay demo script:
  - Confirm ETA display is visibly steadier with `eta_smoothing` enabled.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -433,6 +433,10 @@ Parameters
     Exponential moving average smoothing factor for speed estimates
     (ignored in GUI mode). Ranges from 0 (average speed) to 1
     (current/instantaneous speed) [default: 0.3].
+* eta_smoothing  : float, optional  
+    Exponential moving average smoothing factor for ETA display.
+    Ranges from 0 (disabled) to 1 (current/instantaneous ETA)
+    [default: 0].
 * bar_format  : str, optional  
     Specify a custom bar string formatting. May impact performance.
     [default: '{l_bar}{bar}{r_bar}'], where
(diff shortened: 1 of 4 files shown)


eliotwrobson:
@Steve2457 I've implemented a similar feature on my fork here https://github.com/eliotwrobson/tldm

PR #1764, opened 2026-06-22
perf: inline miniters as local variable in __iter__ loop

AguiPJ:
- Replace self.miniters dict lookup with local variable in hot loop
- Lazy-compute eta datetime only when {eta} is in bar_format
- ~2-3% speedup on tight iteration loops (3M iter test)
- All 107 tests pass


casperdcl (commented):
Users currently can use `format_dict[eta]` even if `'eta' not in bar_format`.

Unsure if a "3% speedup on tight loops" is worth dropping `format_dict[eta]` over.

PR #1767, opened 2026-07-07
Apply disp_trim to all format_meter return paths (#1658)

### LINKED ISSUE #1658: `disp_trim` is not called for all return values of `format_meter`
    + [ ] exception-raising bug
    + [x] visual output bug
  read the [known issues]
  ```python
  import tqdm, sys
  ```

## Versions

- tqdm version: 4.67.1
- Python version: 3.12.7 | packaged by Anaconda, Inc. | (main, Oct  4 2024, 13:17:27) [MSC v.1929 64 bit (AMD64)]
- Operating system: win32

## Facts

The `format_meter` method has multiple return statements:

- https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L618
- https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L633
- https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L643
- https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L652
- https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L657
- https://github.com/tqdm/tqdm/blob/0ed5d7f18fa3153834cbac0aa57e8092b217cc16/tqdm/std.py#L660

2 of them use `disp_trim` and 4 of them do not.

## Problem statement and use case

I often use `bar_format="{desc}"` together with the `set_description_str` method to create status bars. Sometimes the desc is wider than the terminal, but it does not seem to be affected by `ncols` or `dynamic_ncols`.

## Proposed solution

Wrap ALL return values of `format_meter` in `disp_trim`.

## My current workaround

```py
tqdm_bar = tqdm(bar_format="{desc}", dynamic_ncols=True, desc="...")

# Monkeypatch the `format_meter` method of the `tqdm` object.
original_format_meter = tqdm_bar.format_meter

@wraps(original_format_meter)
def format_meter(*args: Any, **kwargs: Any) -> str:
    result = original_format_meter(*args, **kwargs)
    return disp_trim(result, tqdm_bar.ncols or 10)

tqdm_bar.format_meter = format_meter  # type: ignore[method-assign]

with tqdm_bar:
    # Do lots of processing here, and update the description from time to time.
    ...
```

steps-re:
Fixes #1658.

## Bug
`tqdm.std.format_meter` has six `return` paths but only two applied `disp_trim(res, ncols)`. Three others returned untrimmed strings, so `ncols` / `dynamic_ncols` were ignored for a custom `bar_format` without a `{bar}` placeholder (e.g. `bar_format="{desc}"` with `set_description_str`) and for the no-total default stats line. Untrimmed output wider than the terminal overflows and wraps.

## Fix
Wrap those three paths in `disp_trim` using the codebase's existing `disp_trim(x, ncols) if ncols else x` guard, so the `ncols=0` ("no bar" sentinel) and `ncols=None` ("unknown width") cases keep their prior behavior. The intentional `ncols==0` early-return is left untrimmed.

## Verification
- `tqdm(bar_format="{desc}", ncols=20)` with a 60-char description now yields `disp_len == 20` (overflowed before).
- Regression asserts added to `test_format_meter` — fail on pre-fix code, pass with the fix.
- Full suite: **137 passed, 6 skipped** (skips are optional deps). Matches the fix the issue author proposed.

### DIFF
--- a/tests/tests_tqdm.py
+++ b/tests/tests_tqdm.py
@@ -275,6 +275,21 @@ def test_format_meter():
                         bar_format=r'{l_bar}{bar}') == " 20%|" + unich(0x258d) + " "
     assert format_meter(20, 100, 12, ncols=6, rate=8.1,
                         bar_format=r'{bar}|test') == unich(0x258f) + "|test"
+    # ncols must trim bar_format outputs that lack a `{bar}` (issue #1658)
+    long_desc = "x" * 50
+    # total known, custom bar_format without `{bar}`
+    assert format_meter(1, 100, 12, ncols=10,
+                        bar_format="{desc}", prefix=long_desc) == "x" * 10
+    # no total, custom bar_format without `{bar}`
+    assert format_meter(1, None, 12, ncols=10,
+                        bar_format="{desc}", prefix=long_desc) == "x" * 10
+    # no total, no bar_format (default stats line)
+    assert len(format_meter(1, None, 12, ncols=10)) == 10
+    # ncols=0 and ncols=None must not trim these paths
+    assert format_meter(1, 100, 12, ncols=None,
+                        bar_format="{desc}", prefix=long_desc) == long_desc
+    assert format_meter(1, None, 12, ncols=0,
+                        bar_format="{desc}", prefix=long_desc) == long_desc
 
 
 def test_ansi_escape_codes():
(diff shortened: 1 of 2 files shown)


casperdcl:
Thanks for this! Duplicate of & merged with #1717.

Issue #1674, opened 2025-07-14
AttributeError: 'tqdm_rich' object has no attribute '_task_id'


ciaoyizhen:
- [x] I have marked all applicable categories:
    + [ ] exception-raising bug
    + [ ] visual output bug
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [ ] I have searched through the [issue tracker] for duplicates
- [ ] I have mentioned version numbers, operating system and
  environment, where applicable:

```python
import time
import warnings
from tqdm.rich import tqdm

warnings.filterwarnings("ignore")

for i in tqdm(range(100), desc="读取文件夹"):
    for j in tqdm(range(100), desc="读取文件", leave=False):
        time.sleep(0.1)
        
```

[source website]: https://github.com/tqdm/tqdm/
[known issues]: https://github.com/tqdm/tqdm/#faq-and-known-issues
[issue tracker]: https://github.com/tqdm/tqdm/issues?q=

Issue #1391, opened 2022-11-09
tqdm.rich fails when the iterator has no size hint


bilelomrani1:
Consider the following example:
```python
from pathlib import Path
import tqdm.rich

for i in tqdm.rich.tqdm(Path(".").glob("**/*.py")):
    do_something(i)
```

This raises the following exception:
```
...
.venv/lib/python3.9/site-packages/tqdm/rich.py", line 34, in render
    total = int(task.total)
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

In my real use case, the iterator is a custom generator expression that cannot be converted to a kwown-size container.

- [ ] I have marked all applicable categories:
    + [x] exception-raising bug
    + [ ] visual output bug
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [x] I have mentioned version numbers, operating system and
  environment, where applicable:
  ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)

  >> 4.64.1 3.9.15 (main, Oct 17 2022, 03:01:19) 
  [Clang 13.1.6 (clang-1316.0.21.2.5)] darwin
  ```

[source website]: https://github.com/tqdm/tqdm/
[known issues]: https://github.com/tqdm/tqdm/#faq-and-known-issues
[issue tracker]: https://github.com/tqdm/tqdm/issues?q=

PR #1773, opened 2026-07-11
rich: fix TypeError on iterables without a length

vidigoat:
    + [x] exception-raising fix
    + [ ] visual output fix
    + [ ] new feature
- [x] If applicable, I have mentioned the relevant/related issue(s): #1391, #1674

## Bug

`tqdm.rich` crashes on any iterable that has no length:

```python
from tqdm.rich import tqdm

for _ in tqdm(enumerate(range(5))):   # or a generator, Path.glob(...), etc.
    pass
```

```
File "tqdm/rich.py", line 31, in render
    total = int(task.total)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

rich sets a task's `total` to `None` when the wrapped iterable has no `__len__`, but `FractionColumn.render` called `int(task.total)` unconditionally, so the loop dies before the first render. Reported in #1391.

The failure inside `__init__` also leaves the bar half-built (`_prog` set, `_task_id` not), so the subsequent `__del__` -> `close()` -> `display()` raised a second, confusing `AttributeError: 'tqdm_rich' object has no attribute '_task_id'` (#1674) that masked the real error.

## Fix

- `FractionColumn.render`: when `total` is `None`, use the `completed` count for unit selection and print the total as `?`, mirroring rich's own `DownloadColumn`. Sized iterables render exactly as before.
- `display()`: guard on `_task_id` rather than `_prog`, so an interrupted `__init__` cannot cause a secondary `AttributeError` during cleanup.

## Verification

- `tqdm.rich.tqdm(enumerate(range(5)))` now completes without error; `FractionColumn` renders `5/?` for an unknown total and `1.5/? K` with `unit_scale=True`.
- A known total is unchanged: `FractionColumn().render(task)` still yields `3/10` for `completed=3, total=10`.
- Added regression tests to `tests/tests_rich.py` (fail on the current code, pass with the fix).
- `flake8` clean; full suite green locally (`tests_rich.py` 3 passed, `tests_tqdm.py` 70 passed / 1 skipped for missing numpy).

### DIFF
--- a/tests/tests_rich.py
+++ b/tests/tests_rich.py
@@ -1,27 +1,18 @@
 """Test `tqdm.rich`."""
-from contextlib import closing
-from io import StringIO
-
-from pytest import mark
-
-from .tests_tqdm import importorskip
-
-
-def test_rich_import():
-    """Test `tqdm.rich` import"""
-    importorskip('tqdm.rich')
+from pytest import importorskip, mark
 
 
 @mark.filterwarnings("ignore:rich is experimental/alpha:"
                      "tqdm.std.TqdmExperimentalWarning")
-def test_rich_no_total():
+def test_rich_no_total(capsys):
     """Test `tqdm.rich` on an iterable without a length, e.g. a generator"""
     rich = importorskip('tqdm.rich')
-    with closing(StringIO()) as our_file:
-        # a generator has no __len__, so the rich task total stays None
-        with rich.tqdm((i for i in range(5)), file=our_file) as pbar:
-            for _ in pbar:
-                pass
+    with rich.tqdm(i for i in range(5)) as pbar:
+        for _ in pbar:
+            pass
+    out, err = capsys.readouterr()
+    assert not err
+    assert '5/?' in out
 
 
 def test_rich_fraction_column_no_total():
--- a/tests/tests_rich.py
+++ b/tests/tests_rich.py
@@ -1,7 +1,41 @@
 """Test `tqdm.rich`."""
+from contextlib import closing
+from io import StringIO
+
+from pytest import mark
+
 from .tests_tqdm import importorskip
 
 
 def test_rich_import():
     """Test `tqdm.rich` import"""
     importorskip('tqdm.rich')
+
+
+@mark.filterwarnings("ignore:rich is experimental/alpha:"
+                     "tqdm.std.TqdmExperimentalWarning")
+def test_rich_no_total():
+    """Test `tqdm.rich` on an iterable without a length, e.g. a generator"""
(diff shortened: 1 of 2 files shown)


tests/tests_rich.py
```diff
@@ -1,7 +1,41 @@
 """Test `tqdm.rich`."""
+from contextlib import closing
+from io import StringIO
+
+from pytest import mark
+
 from .tests_tqdm import importorskip
```
casperdcl:
```suggestion
from pytest import importorskip, mark
```

tqdm/rich.py
```diff
@@ -124,7 +127,7 @@ def clear(self, *_, **__):
         pass
 
     def display(self, *_, **__):
-        if not hasattr(self, '_prog'):
+        if not hasattr(self, '_task_id'):
```
casperdcl, line 130: is this change needed?

tests/tests_rich.py
```diff
@@ -1,7 +1,41 @@
...
+@mark.filterwarnings("ignore:rich is experimental/alpha:"
+                     "tqdm.std.TqdmExperimentalWarning")
+def test_rich_no_total():
+    """Test `tqdm.rich` on an iterable without a length, e.g. a generator"""
+    rich = importorskip('tqdm.rich')
+    with closing(StringIO()) as our_file:
+        # a generator has no __len__, so the rich task total stays None
+        with rich.tqdm((i for i in range(5)), file=our_file) as pbar:
+            for _ in pbar:
+                pass
```
casperdcl:
```suggestion
def test_rich_no_total(capsys):
    """Test `tqdm.rich` on an iterable without a length, e.g. a generator"""
    rich = importorskip('tqdm.rich')
    with rich.tqdm((i for i in range(5))) as pbar:
        for _ in pbar:
            pass
    out, err = capsys.readouterr()
    assert not err
    assert '5/?' in out
```

tests/tests_rich.py
```diff
@@ -1,7 +1,41 @@
...
+from pytest import mark
+
 from .tests_tqdm import importorskip
 
 
 def test_rich_import():
     """Test `tqdm.rich` import"""
     importorskip('tqdm.rich')
+
+
```
casperdcl:
```suggestion
```

casperdcl (commented):
Thanks! fixes #1674, fixes #1391

tqdm/rich.py
casperdcl: ah I see the commit commit comment (fixes #1674)

PR #1777, opened 2026-07-17
add `tqdm.concurrent.intepreter_map`

shermansiu:
Python 3.14 adds an [InterpreterPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.InterpreterPoolExecutor) - this PR adds `interpreter_map` that uses it. The `InterpreterPoolExecutor` import is lazy so that `tqdm.concurrent.futures` can still be imported in older Python versions.

### DIFF
--- a/tests/tests_concurrent.py
+++ b/tests/tests_concurrent.py
@@ -13,6 +13,34 @@ def incr(x):
     return x + 1
 
 
+def check_lock(args):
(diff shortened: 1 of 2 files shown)


casperdcl:
Thanks for this! (un)fortunately just merged in several related PRs so there are conflicts now.

casperdcl (approved):
Thanks for this!

Issue #1537, opened 2023-12-13
AttributeError: 'tqdm' object has no attribute 'last_print_t' on Python 3.12


mweinelt:
- [x] I have marked all applicable categories:
    + [x] exception-raising bug
    + [ ] visual output bug
- [x] I have visited the [source website], and in particular
  read the [known issues]
- [x] I have searched through the [issue tracker] for duplicates
- [x] I have mentioned version numbers, operating system and
  environment, where applicable:
  ```python
  import tqdm, sys
  print(tqdm.__version__, sys.version, sys.platform)
  ```
[source website]: https://github.com/tqdm/tqdm/
[known issues]: https://github.com/tqdm/tqdm/#faq-and-known-issues
[issue tracker]: https://github.com/tqdm/tqdm/issues?q=

### Environment

Python 3.12.1
pytest 7.4.3
pytest-asyncio 0.23.2
pluggy 1.3.0
tqdm 4.66.1

### Description

I'm aware Python 3.12 is not supported according to the trove classifiers, but I hope it will be soon. We're seeing the following exception many times during testing on Python 3.12.1, the tests run fine on 3.11.7.

```pytb
Exception ignored in: <function tqdm.__del__ at 0x7ffff64a2c00>
Traceback (most recent call last):
  File "/build/tqdm-4.66.1/tqdm/std.py", line 1149, in __del__
    self.close()
  File "/build/tqdm-4.66.1/tqdm/std.py", line 1278, in close
    if self.last_print_t < self.start_t + self.delay:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'tqdm' object has no attribute 'last_print_t'
```

stanislavkozlovski:
I'm hitting this with Python **3.11.8** and latest tqdm **4.66.5**

maurerle:
We were hitting this in a setting on windows where an error happened when creating the `tqdm` as the stderr stream was redirected to something else.

As an error was thrown, it did not fully initialize, throwing this error.

Something like this example will throw the error on windows for example:
```python
@contextlib.contextmanager
def suppress_output():
    # Save the original stdout and stderr file descriptors
    original_stdout_fd = sys.stdout.fileno()
    original_stderr_fd = sys.stderr.fileno()

    # Open /dev/null for redirecting output
    devnull = os.open(os.devnull, os.O_WRONLY)

    # Duplicate the original stdout and stderr file descriptors to restore them later
[...]

Issue #261, opened 2016-09-08
AttributeError: 'tqdm' object has no attribute 'last_print_t'


chenasraf:
Didn't use to happen before, not sure why it started happening. I've got 
```
Traceback (most recent call last):
  File "main.py", line 206, in <module>
    OrgbaseQ(**settings).start()
  File "/Users/casraf/Dev/whitespace-tools/item_q/item_q/item_q.py", line 26, in start
  File "main.py", line 125, in flush
    tqdm.write('Found %d matching people by names' % pg_count)
  File "/usr/local/lib/python2.7/site-packages/tqdm/_tqdm.py", line 343, in write
    inst.refresh()
  File "/usr/local/lib/python2.7/site-packages/tqdm/_tqdm.py", line 944, in refresh
    self.fp.write(self.__repr__())
  File "/usr/local/lib/python2.7/site-packages/tqdm/_tqdm.py", line 649, in __repr__
    time() - self.last_print_t,
AttributeError: 'tqdm' object has no attribute 'last_print_t'
```

CrazyPython:
@chenasraf surrounding code?

chenasraf:
```python
            pg_cursor = self.pgdb.cursor()
            pg_query = """
                SELECT id, first_name, last_name, emails #>> '{{0,address}}' AS email, jobs
                FROM people
                WHERE (jobs IS NOT NULL OR emails #>> '0' IS NOT NULL)
                AND ({conditions})
            """.format(conditions=' OR '.join(conditions))
            count = 0

            tqdm.write('Found %d matching people by names' % self._pg_count_query(pg_cursor, pg_query))  # error points here
            pg_cursor.execute(pg_query)
            for person in tqdm(pg_cursor.fetchall(), **self.options['tqdm_defaults']):
                pg_id, first_name, last_name, email, jobs = person
```

I have to point out I have about 2-3 progress bar working alongside each other here, not sure if that has anything to do with it but I'm guessing it does.

lrq3000:
Hello @chenasraf, are the bars working in parallel or sequentially? If they are working in parallel, then you may be running into a racing condition, the write method is indeed not parallel safe. But it may be fixed if that's the issue here.

lrq3000:
@chenasraf If your bars are indeed working in parallel, please try this experimental release and tell us if this fixes your issue:

https://github.com/tqdm/tqdm/archive/small-fix-refresh.zip

Thank's!

lrq3000:
[...]

PR #1778, opened 2026-07-18
Fix AttributeError in close() when __init__ fails before last_print_t is set

### LINKED ISSUE #1668: version 4.67.1 bug
Exception ignored in: <function tqdm.__del__ at 0x15d7ab920>
  File "/opt/anaconda3/envs/xiaozhi/lib/python3.11/site-packages/tqdm/std.py", line 1148, in __del__
    self.close()
  File "/opt/anaconda3/envs/xiaozhi/lib/python3.11/site-packages/tqdm/std.py", line 1277, in close
    if self.last_print_t < self.start_t + self.delay:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'tqdm_asyncio' object has no attribute 'last_print_t'

deeferentleeg:
Fixes #1668 (also relates to #1537, #261).

## Problem

When `tqdm.__init__` fails partway through (e.g. `self.refresh()` at the end of `__init__` raises an I/O error) and the object is subsequently garbage-collected, `__del__` calls `self.close()`, which accesses `self.last_print_t` — an attribute set only at the very end of `__init__`. If `__init__` did not reach that point, `close()` raises:

```
AttributeError: 'tqdm_asyncio' object has no attribute 'last_print_t'
```

This is especially visible with `tqdm_asyncio` (which inherits `close()` from `tqdm.std.tqdm`) when the asyncio event loop or file handle is in an unexpected state during construction.

## Fix

Add two defensive guards in `close()`:

1. Replace `if self.disable:` with `if getattr(self, 'disable', True):` — handles the case where `__init__` failed so early that `disable` was never set. Defaults to `True` (treat as already-closed → return).

2. Add `if not hasattr(self, 'last_print_t'): return` before the `self.last_print_t < ...` comparison — handles the case where `__init__` progressed past the `disable` assignment but failed before the time-counter initialisation.

Both guards are pure safety-nets with **zero behaviour change on the happy path** (fully-initialised instances always have both attributes). The `hasattr` pattern is already used in `display()` (line 1496: `if not hasattr(self, "sp")`), so this is consistent with existing conventions.

## Testing

- Reproduced the original crash with a mock object simulating partial `__init__`.
- Verified the patched `close()` handles partial init, very-early init failure, and the normal happy path without regression.
- `python -m py_compile tqdm/std.py` passes.

### DIFF
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -1270,8 +1270,6 @@ class tqdm(Comparable):
 
     def close(self):
         """Cleanup and (if leave=False) close the progress bar."""
-        # guard against partially-initialised instances (e.g. __init__ failed,
-        # or __del__ triggered by GC before init completed). See GH #1668.
         if getattr(self, 'disable', True):
             return
 
@@ -1283,7 +1281,6 @@ class tqdm(Comparable):
         self._decr_instances(self)
 
         if not hasattr(self, 'last_print_t'):
-            # __init__ did not reach the time-counter initialisation
             return
         if self.last_print_t < self.start_t + self.delay:
             # haven't ever displayed; nothing to clear
--- a/tqdm/std.py
+++ b/tqdm/std.py
@@ -1270,7 +1270,9 @@ class tqdm(Comparable):
 
     def close(self):
         """Cleanup and (if leave=False) close the progress bar."""
-        if self.disable:
+        # guard against partially-initialised instances (e.g. __init__ failed,
+        # or __del__ triggered by GC before init completed). See GH #1668.
+        if getattr(self, 'disable', True):
             return
 
         # Prevent multiple closures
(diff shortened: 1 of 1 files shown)


tqdm/std.py
```diff
@@ -1280,6 +1282,9 @@ def close(self):
         pos = abs(self.pos)
         self._decr_instances(self)
 
+        if not hasattr(self, 'last_print_t'):
+            # __init__ did not reach the time-counter initialisation
```
```suggestion
```

tqdm/std.py
```diff
@@ -1270,7 +1270,9 @@ def update(self, n=1):
 
     def close(self):
         """Cleanup and (if leave=False) close the progress bar."""
-        if self.disable:
+        # guard against partially-initialised instances (e.g. __init__ failed,
+        # or __del__ triggered by GC before init completed). See GH #1668.
```
```suggestion
```

casperdcl (approved):
Thanks!

PR #1780, opened 2026-07-22
docs: Fix word order 'can be also used' -> 'can also be used'

LaRoyBot:
Fixes awkward word order ("can be also used" -> "can also be used") in the main README.


casperdcl (approved):
Thanks!

PR #1783, opened 2026-07-27
Treat inf total as unknown in reset() too

LuShadowX:
Follow-up to #1781 / #651.

`__init__` normalises an infinite total to `None` ("Infinite iterations, behave same as unknown"), but `reset()` assigns `total` straight through, so the same value takes a different path:

```python
tqdm(total=float("inf")).total   # None

t = tqdm(total=10)
t.reset(total=float("inf"))
t.total                          # inf
len(t)                           # TypeError: 'float' object cannot be interpreted as an integer
bool(t)                          # True   (vs TypeError for tqdm(total=inf))
t.format_dict['total']           # inf    (vs None)
```

That `len()` TypeError is the one #651 set out to remove — the display itself is fine since #1781, but the state behind it isn't.

`reset()` now applies the same normalisation as `__init__`. `notebook`, `rich` and `tk` all delegate to `super().reset(total=total)`, so they pick this up as well.

One thing I left alone deliberately: those three subclasses configure their widgets from the raw `total` before calling `super()`, so e.g. `tqdm.tk` still sets `maximum=inf, mode="determinate"` for `reset(total=inf)`. Happy to extend the fix there if you'd like it in the same PR.

Added a test — fails on master with `assert inf is None`, passes with the fix. Full suite: 151 passed, 7 skipped.

### DIFF
--- a/tests/tests_tqdm.py
+++ b/tests/tests_tqdm.py
@@ -1144,6 +1144,19 @@ def test_reset():
         assert '| 10/12' in our_file.getvalue()
 
 
+def test_reset_inf():
+    """Test resetting a bar to an infinite total (#651)"""
+    with closing(StringIO()) as our_file:
+        with tqdm(total=10, file=our_file,
+                  miniters=1, mininterval=0, maxinterval=0) as t:
+            t.update(5)
+            t.reset(total=float("inf"))
+            t.update()
+            # same as tqdm(total=float("inf")): treated as unknown
+            assert t.total is None
+        assert '1it' in our_file.getvalue()
+
+
 def test_disabled_reset(capsys):
     """Test disabled reset"""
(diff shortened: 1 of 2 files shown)


casperdcl (approved):
Thanks!

LuShadowX:
happy to help if u need more feel free to ask

casperdcl:
well, might be worth checking & fixing subclass behaviour too.

LuShadowX:
Good call — the rich one was worse than I thought (`reset()` raised `TypeError` for any total). Fixed all three in #1784.

LuShadowX:
Thanks for the merge too! If there's anything else in tqdm you'd like a second pair of hands on, feel free to ping me.

PR #1788, opened 2026-08-10
Fix concurrent maps for unsized iterables

nyxst4ck:
## Summary

Allow `thread_map()` and `process_map()` to accept iterables that do not implement `__len__` or `__length_hint__`.

## Problem and reproduction

`_min_map_len()` currently calls `min()` over only the iterables with a known length. When every input is unsized, the filtered generator is empty and `min()` raises `ValueError` before the executor starts:

```python
thread_map(abs, (value for value in [-1, -2]))
```

Return `None` when no length is known, and only perform the large-iterable `chunksize` warning check when a length is available. Mixed sized/unsized inputs continue to use the shortest known length.

## Tests

- `python -m pytest tests/tests_concurrent.py -q`
- Result: 14 passed, 2 skipped (Python 3.14-only tests)
- `git diff --check upstream/master...HEAD`

### DIFF
--- a/tests/tests_concurrent.py
+++ b/tests/tests_concurrent.py
@@ -52,6 +52,17 @@ def test_process_map():
             skip(str(err))
 
 
+def test_thread_map_unsized():
+    """Test contrib.concurrent.thread_map with unsized iterables"""
+    with closing(StringIO()) as our_file:
+        a = range(9)
+        b = [i + 1 for i in a]
+        # no iterable reports a length
+        assert thread_map(incr, (i for i in a), file=our_file) == b
+        # only some iterables report a length
+        assert thread_map(lambda x, _: x + 1, (i for i in a), a, file=our_file) == b
+
+
 def check_lock(args):
     """Check that another interpreter cannot acquire a held tqdm lock"""
     from os.path import exists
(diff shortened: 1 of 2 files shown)


MohammedAlkindi (commented):
Reproduced this and the fix holds. Windows 11, Python 3.13.13.

`thread_map` over inputs with no `__len__` and no `__length_hint__`:

```
                        master                                  this PR
sized only (control)    [0, 1, 4, 9, 16]                        [0, 1, 4, 9, 16]
unsized only            ValueError: min() iterable argument     [0, 1, 4, 9, 16]
                        is empty
mixed sized + unsized   [0, 1, 4, 9, 16]                        [0, 1, 4, 9, 16]
```

The mixed row is the one worth having in the control set: it already worked on `master`, because one sized input is enough to keep the generator non-empty, so the bug only appears when **every** input is unsized. That is a narrower trigger than the title suggests and it is worth someone knowing when they read the fix.

Full suite, and this branch is based on current `master` so the numbers compare directly:

```
master    149 passed, 11 skipped
this PR   153 passed, 11 skipped
```

Zero failures either side. Of the four new tests, **three fail on unmodified source** with exactly `ValueError: min() iterable argument is empty`, checked by restoring only `tqdm/contrib/concurrent.py` from `master` and keeping the tests. The fourth passes on `master` too, which is right: it is the guard proving the sized path did not change.

I went looking for consumers of `_min_map_len` that would now receive `None` and choke, since that is the usual way a `default=` fix leaks. There are two and both are fine:

- line 127, `kwargs['total'] = _min_map_len(iterables)`. `total=None` is tqdm's own idiom for an unknown length, so this is the correct value rather than a papered-over one, and it is why the bar still renders for the unsized case.
- line 252, guarded with `is not None` in this PR before the `> 1000` comparison.

So the change is complete for this module, not one instance of a pattern.

Not a maintainer, just reporting what ran here.

MohammedAlkindi:
Reproduces and fixes cleanly on Windows 11, Python 3.13.13.

On `master` 96f2e60, `thread_map(abs, (v for v in [-1, -2]))` raises `ValueError: min() iterable argument is empty`. On `d28f76a` it returns, and the bar renders `3it [00:00, 6135.01it/s]` for a 3-item generator: right count, no total, which is what an unsized input should give.

Both claims in the description hold. Mixed sized and unsized still takes the shortest known length, `range(5)` against a 100-item generator giving `5/5`. Sized-only is unchanged at `7/7`. An explicit `total=` on an unsized iterable still works.

`tests/tests_concurrent.py`: 10 passed, 2 skipped on `master`; 14 passed, 2 skipped here, matching your numbers.

PR #1792, opened 2026-08-14
fix: count process_map items when chunksize > 1

### LINKED ISSUE #1791: process_map undercounts progress with chunksize > 1 in tqdm 4.70.0
    + [ ] exception-raising bug
    + [X] visual output bug
  read the [known issues]
```text
Fedora 43
tqdm 4.70.0
Python 3.14.6 (main, Jun 11 2026, 00:00:00)
[GCC 15.2.1 20260123 (Red Hat 15.2.1-7)]
linux
```

I observed a progress-bar regression after upgrading from tqdm 4.69.1 to 4.70.0.

Here's an example:

```python
def f(x): return x
if __name__ == "__main__": process_map(f, range(101), chunksize=10)
```

4.69.1 finishes correctly at 101/101
4.70.0 finishes **incorrectly** at **11**/101.

I asked ChatGPT to investigate. It reproduced the issue on Python 3.13.5 and traced it to [PR #1709](https://github.com/tqdm/tqdm/pull/1709), specifically this new callback:

```python
fut.add_done_callback(lambda _: pbar.update())
```

With `process_map(..., chunksize=10)`, each Future represents a chunk of up to 10 input items. The callback therefore increments the bar once per chunk, while the total remains the number of items. For 101 items, `ProcessPoolExecutor` creates 11 chunks, so the bar ends at 11/101.

I can provide additional test details if that will help.

DSeaStar:
Fixes #1791

    + [ ] exception-raising fix
    + [x] visual output fix
    + [ ] new feature

`ProcessPoolExecutor.map` submits one Future per *chunk*. After #1709, `process_map` called `pbar.update()` once per Future, so `chunksize=10` on 101 items finished at **11/101** instead of 101/101.

This updates the bar by the chunk length (last chunk included). `thread_map` / `interpreter_map` still increment once per item.

```python
process_map(lambda x: x, range(101), chunksize=10)
# 4.70.0: 11/101
# this PR: 101/101
```

Regression test: `test_process_map_chunksize_progress`.

### DIFF
--- a/tests/tests_concurrent.py
+++ b/tests/tests_concurrent.py
@@ -52,6 +52,18 @@ def test_process_map():
             skip(str(err))
 
 
+def test_process_map_chunksize_progress():
+    """process_map should count items, not chunks, when chunksize > 1 (#1791)"""
+    with closing(StringIO()) as our_file:
+        a = range(101)
+        b = [i + 1 for i in a]
+        try:
+            assert process_map(incr, a, chunksize=10, file=our_file, miniters=1) == b
+        except ImportError as err:
+            skip(str(err))
+        assert '101/101' in our_file.getvalue()
+
+
 def check_lock(args):
     """Check that another interpreter cannot acquire a held tqdm lock"""
     from os.path import exists
--- a/tqdm/contrib/concurrent.py
+++ b/tqdm/contrib/concurrent.py
@@ -154,10 +154,14 @@ def _executor_map(
                 if dynamic_miniters is not None:
                     pbar.dynamic_miniters = True
                 orisubmit = ex.submit
+                # ProcessPoolExecutor.map submits one Future per chunk, not per item.
+                from concurrent.futures import ProcessPoolExecutor
+                counts_chunks = PoolExecutor is ProcessPoolExecutor
 
                 def patchsubmit(*args, **kwargs):
                     fut = orisubmit(*args, **kwargs)
-                    fut.add_done_callback(lambda _: pbar.update())
+                    n = len(args[1]) if counts_chunks and len(args) > 1 else 1
+                    fut.add_done_callback(lambda _, n=n: pbar.update(n))
                     return fut
                 ex.submit = patchsubmit
                 return list(ex.map(


MohammedAlkindi (commented):
Ran this on Windows 11 (10.0.26200), Python 3.13.13, against `upstream/master` at `96f2e60`.

**Test evidence (delta against base):** `tests/tests_concurrent.py` is **10 passed / 2 skipped** on master and **11 passed / 2 skipped** on this branch — the +1 is this PR's own `test_process_map_chunksize_progress`. No regressions. (The 2 skips are the `requires Python 3.14+` interpreter-pool tests, skipped on both sides here.)

**Heads-up: #1793 fixes the same issue.** It also targets #1791, adds a near-identical test (`test_process_map_chunksize`, same `range(101)` / `chunksize=10` / `'101/101'` assertion), and patches the same callback. Flagging it because the two PRs were opened within a day of each other and only one can land — not to argue for either.

**The one substantive difference between them is the executor test:**

```python
# this PR (#1792)
counts_chunks = PoolExecutor is ProcessPoolExecutor

# #1793
if chunksize > 1 and not issubclass(PoolExecutor, ThreadPoolExecutor):
```

`is` is an identity check, so it returns False for any *subclass* of `ProcessPoolExecutor` and such a caller would silently keep the old per-chunk undercount. `issubclass` doesn't have that hole. In fairness this is currently unreachable through the public API — `process_map` always passes `ProcessPoolExecutor` itself and `thread_map` passes `ThreadPoolExecutor` — so it's robustness, not a live bug.

**I checked both against the interpreter pool merged in #1777, since that's the newest executor and the relevant tests skip below 3.14.** On Python 3.14.4:

```
InterpreterPoolExecutor.__mro__[:3] -> InterpreterPoolExecutor, ThreadPoolExecutor, Executor
issubclass(InterpreterPoolExecutor, ThreadPoolExecutor) -> True
InterpreterPoolExecutor is ProcessPoolExecutor            -> False
```

So both approaches leave `interpreter_map` at `pbar.update(1)` per future, which is correct — `InterpreterPoolExecutor` inherits `ThreadPoolExecutor.map` and ignores `chunksize`. Neither PR regresses #1777.

**One small thing in this PR specifically:** the `from concurrent.futures import ProcessPoolExecutor` is inside `_executor_map`, executed on every call. It's cached in `sys.modules` so the cost is negligible, but the module already imports from `concurrent.futures` at top level (`length_hint` aside), so a module-level import would match the surrounding style — unless it's deliberately local to dodge a circular import I haven't spotted.

---

*Disclosure: I used an AI assistant while reviewing. The suite runs, the 3.14 MRO check, and the comparison against #1793 above were all executed by me on the machine described.*

PR #1793, opened 2026-08-14
fix(contrib): count chunk size for process_map progress (#1791)

### LINKED ISSUE #1791: process_map undercounts progress with chunksize > 1 in tqdm 4.70.0

alexchen-sys:
    + [x] visual output fix

Fixes #1791.

With `chunksize > 1`, `process_map` submits one future per chunk to the
worker pool, but the done callback only advanced the progress bar by one
item per future — so the bar stopped well short of the total (e.g.
`11/101` with `chunksize=10`). The callback now advances the bar by the
exact size of each submitted chunk. `ThreadPoolExecutor` (and its
subclasses) ignore `chunksize` and submit one future per item, so that
path keeps the previous one-item-per-future behaviour.

Tests added for `process_map`, `thread_map` and `interpreter_map`
(pending Python 3.14) with `chunksize > 1`.

  read the [known issues]
  ```python
  import tqdm, sys
  ```
  4.67.3 3.11.15 linux

### DIFF
--- a/tests/tests_concurrent.py
+++ b/tests/tests_concurrent.py
@@ -52,6 +52,43 @@ def test_process_map():
             skip(str(err))
 
 
+def test_process_map_chunksize():
+    """Test contrib.concurrent.process_map progress with chunksize > 1 (#1791)"""
+    with closing(StringIO()) as our_file:
+        a = range(101)
+        b = [i + 1 for i in a]
+        try:
+            assert process_map(incr, a, chunksize=10, max_workers=2,
+                               file=our_file) == b
+        except ImportError as err:
+            skip(str(err))
+        # each submitted future processes a whole chunk of items, so each
+        # chunk completion must advance the bar by more than one
+        assert '101/101' in our_file.getvalue()
+
+
+def test_thread_map_chunksize():
+    """Test contrib.concurrent.thread_map progress with chunksize > 1 (#1791)"""
+    with closing(StringIO()) as our_file:
+        a = list(range(17))
+        b = [i + 1 for i in a]
+        assert thread_map(incr, a, chunksize=4, file=our_file) == b
+        assert '17/17' in our_file.getvalue()
+
+
+@mark.skipif(sys.version_info < (3, 14), reason="requires Python 3.14+")
+def test_interpreter_map_chunksize(capsys):
+    """Test contrib.concurrent.interpreter_map progress with chunksize > 1 (#1791)"""
+    a = range(101)
+    b = [i + 1 for i in a]
+    try:
+        assert interpreter_map(incr, a, chunksize=10) == b
+    except ImportError as err:
+        skip(str(err))
+    _, err = capsys.readouterr()
+    assert '101/101' in err
+
+
 def check_lock(args):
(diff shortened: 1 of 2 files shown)


MohammedAlkindi (commented):
Reproduced #1791 independently and this fixes it. Windows 11, `range(101)`, `max_workers=2`, reading the final count off the rendered bar.

<details>
<summary>probe, master vs this branch, on both interpreters</summary>

```
                                    master        this PR
process_map chunksize=1  (control)  101/101       101/101
process_map chunksize=10            11/101        101/101
thread_map  chunksize=10            101/101       101/101
```

Same result on 3.13.13 and 3.14.7. `thread_map` was already correct because `Executor.map` ignores `chunksize` and submits one future per item, which is exactly the asymmetry the new guard encodes.

`tests/tests_concurrent.py`:

```
3.13.13   master 10 passed, 2 skipped   ->  PR 12 passed, 3 skipped
3.14.7    master 12 passed, 0 skipped   ->  PR 15 passed, 0 skipped
```
</details>

The three skips on 3.13 are the `requires Python 3.14+` gates, so I ran the whole file on 3.14 as well to make sure `interpreter_map` was actually exercised rather than skipped past. It passes.

Worth stating explicitly since #1777 landed recently: the guard does the right thing for `interpreter_map` by construction, not by luck.

```
issubclass(InterpreterPoolExecutor, ThreadPoolExecutor)  -> True
InterpreterPoolExecutor.map is Executor.map              -> True
```

So it takes the per-item branch and its behaviour is unchanged by this PR.

One robustness note, take it or leave it. `n=len(args[1])` reads the second positional of `Executor.map`'s internal `self.submit(fn, *args)`, so it assumes the chunked call shape. `_executor_map` accepts an arbitrary `PoolExecutor`, and a custom one that is not a `ThreadPoolExecutor` subclass but whose `map` submits one item per future would hit `len(<item>)` at submit time. A `getattr` with a fallback of 1, or a comment naming the assumption, would make that harmless.

PR #1835, opened 2026-09-14
std: don't acquire lock for disabled bars

### LINKED ISSUE #1500: Don't attempt to acquire global lock when disable=True
    + [ ] exception-raising bug
    + [ ] visual output bug
  read the [known issues]
  ```python
  import tqdm, sys
  # 4.65.0 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:25:29) [Clang 14.0.6 ] darwin
  ```

I ran a program that uses `multiprocessing` and `tqdm`; however, `tqdm` was always initialized with `disable=True`. When one of the processes exited without cleaning up properly, another of the processes ended up hanging trying to acquire the `tqdm` lock. I was surprised that `tqdm` creates and uses a shared cross-process lock even if `disable` is always set to `True`. Because shared cross-process locks are potentially dangerous, consider skipping creating/acquiring the lock when initialized with `disable=True`.

anandghegde:
    + [x] exception-raising fix
    + [ ] visual output fix
    + [ ] new feature
- [x] Any AI-assisted commits are clearly marked ([author], [Co-authored-by] or [Assisted-by]) using the syntax `Provider Model <email>`, e.g. `Co-authored-by: Claude Opus 5 <noreply@anthropic.com>`

Fixes #1500

A bar created with `disable=True` still takes tqdm's shared write lock twice: once in `__new__` (to add itself to `_instances` and start the monitor thread) and again in `__init__` (to take itself back out). If that lock is a `multiprocessing.RLock` and a process died while holding it, every later `tqdm(..., disable=True)` blocks forever. So a library that only ever creates disabled bars can still hang.

Repro on master (fork start method):

```python
import multiprocessing as mp, os, threading
from tqdm import tqdm

def die_holding_lock():
    tqdm.get_lock().acquire()
    os._exit(1)

if __name__ == "__main__":
    mp.set_start_method("fork")
    tqdm.get_lock()
    p = mp.Process(target=die_holding_lock); p.start(); p.join()
    done = threading.Event()
    threading.Thread(target=lambda: (list(tqdm(range(3), disable=True)), done.set()), daemon=True).start()
    print("ok" if done.wait(3) else "HANG")
```

master prints `HANG`; this branch prints `ok`.

Changes:
- Registering in `_instances` and starting the monitor move from `__new__` into `__init__`, inside the lock block that already sets `pos`. That block only runs for enabled bars, so disabled bars never touch the lock, and they no longer start a `TMonitor` thread either.
- The disabled and unknown-kwargs paths used to add then remove the instance under the lock. They now skip registration. A disabled bar's `pos` is `0`, or `-position` if one was given. It isn't shown, and it isn't counted when other bars pick positions.
- Added `test_disable_lock`, which swaps in a lock that raises on `acquire`. It fails on master and passes here.

Tests: `pytest -k "not perf"` passes, 175 passed and 5 skipped (tk isn't available locally), with numpy/pandas/rich/dask/matplotlib installed. `pre-commit` flake8 is clean on `tqdm/std.py`. Checked on macOS, Python 3.12.

The code was written with Claude Code. I reviewed it and ran the repro and tests above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01MA8JGs7pCoXFL2W6jEJ8sD


anandghegde:
The failing `pre-commit.ci - pr` check here is not caused by this PR — `flake8` fails the same way on a clean `master`.

```
benchmarks/benchmarks.py:20:9: B018 Found useless Call expression. ...
tests/tests_tqdm.py:1219:9: B018 Found useless Call expression. ...
```

Neither line is in this diff. `tests/tests_tqdm.py:1219` is `str(t)` inside the pre-existing `test_disabled_repr`; it only moved from 1201 to 1219 because this PR adds 18 lines above it. The file is byte-identical to `master` in that region.

Cause: `.pre-commit-config.yaml` pins `flake8` to 7.3.0 but leaves `flake8-bugbear` unpinned, and `B018` was extended to cover `Call` expressions in the 26.x line. Bisected against `master` at 9cf5a12:

| flake8-bugbear | result |
| --- | --- |
| 25.11.29 | clean, exit 0 |
| 26.9.9 | the two `B018`s above, exit 1 |

```console
$ git worktree add /tmp/tqdm-master 9cf5a12 && cd /tmp/tqdm-master
$ uvx --from flake8==7.3.0 --with flake8-pyproject --with 'flake8-bugbear<26' flake8 -j8 --select=B018 .   # exit 0
$ uvx --from flake8==7.3.0 --with flake8-pyproject --with flake8-bugbear     flake8 -j8 --select=B018 .   # exit 1
```

So every open PR against `master` is red right now, not just this one.

Both flagged expressions look deliberate — `all(pbar)` is the benchmark, and `str(t)` is exercising `__str__` on a disabled bar — so `# noqa: B018` on the two lines seems more honest than rewriting them (`benchmarks.py:20` already carries a `# pylint: disable=pointless-statement` saying the same thing). Happy to open that as a separate one-line-each PR so this one can be judged on its own diff; just say the word, or pin `flake8-bugbear<26` instead if you'd rather not sprinkle `noqa`s.

